PR #1, opened 2016-01-29
Fixed typos in readme

gshively11:


gshively11:
Just noticed that the .bin links are `godaddy-js-style` still, should we update those to match the package name @indexzero?

indexzero:
This is not a typo. godaddy-style is the package name on npm. https://github.com/godaddy/javascript/blob/master/package.json

indexzero:
My brain swapped this diff. Nvm.

PR #5, opened 2016-02-15
[major] update dependencies and eslint config

Swaagie:
- Updates the config to reflect changes here http://eslint.org/docs/user-guide/migrating-to-2.0.0
- Update dependencies to latest where possible

What would be a solid way to test the usage of each plugin with the .eslintrc config file?


gshively11:
This doesn't look like it will address the issue with ES5 projects and global `"use strict";` receiving errors, due to ES6's default strict mode and eslint's rule to disallow unnecessary global strict flags.  Are there any recommendations for this situation?

The other changes in this PR look solid, I pulled it down in my project and it ran OK, with the exception of the "use strict;" errors mentioned above.

:+1:

gshively11:
As far as how to test each plugin, I'm not sure.  I haven't used eslint until recently, so I'm still fairly unfamiliar with its behavior/construction.

Swaagie:
Yeah I've been running into the same issue, but since I'm mostly writing es6 atm I didn't care enough for it. But I agree that this should be fixed

gshively11:
I can take the initiative on that one (will help me learn more about fashion-show and eslint).  I can't use godaddy-style until it's resolved.

Swaagie:
As reference https://github.com/eslint/eslint/issues/2785 seems to have been reported before. Do you want access to my fork or will you create a separate PR?

gshively11:
If I'm reading that long chain of issues correctly, there is no way to disable those warnings if you have `ecmaFeatures.modules: true`.  So we'll probably need a way for apps consuming godaddy-style to call out their ES5/ES6 status?

gshively11:
I'll work in a separate PR for this one, once we finish discussions.

gshively11:
It looks like removing `"parser": "babel-eslint"` from .eslintrc prevents the "use strict" warnings

gshively11:
Actually, looking again at this PR, why aren't all the changes to the `dotfiles/.eslintrc` made in `dist/.eslintrc` as well?  The `dist` version is the one used in godaddy-style, yea?

hzoo:
You can update babel-eslint to 4.1.8, although it's not updated for eslint 2 atm. https://github.com/babel/babel-eslint/issues/254

Swaagie:
My bad, sorry hadn't run `prepublish`, generated files should be updated now as well

gshively11:
Ok, re-:+1:.  Thoughts on having godaddy-style generate an `.eslintrc-es5` that doesn't use babel-eslint, and provide an argument flag in godaddy-js-style-eslint that uses the es5 version for the configDir option in fashion-show?  I don't see another option for ES5 projects to use godaddy-style until babel-eslint honors the correct parserOptions.

Swaagie:
@indexzero thoughts on ^?

indexzero:
@gshively11 I am +1 on this, but probably something like `dist/es5/.eslintrc`

> Thoughts on having godaddy-style generate an `.eslintrc-es5` that doesn't use babel-eslint, and provide an argument flag in godaddy-js-style-eslint that uses the es5 version for the configDir option in fashion-show? I don't see another option for ES5 projects to use godaddy-style until babel-eslint honors the correct parserOptions.

package.json
```diff
@@ -25,21 +25,21 @@
...
-    "eslint-plugin-json": "^1.1.0"
+    "eslint-plugin-mocha": "^2.0.0",
+    "eslint-plugin-react": "^3.16.1",
+    "eslint-plugin-json": "^1.2.0"
   },
   "dependencies": {
     "babel-eslint": "^4.1.3",
-    "escope": "3.3.0",
-    "eslint": "^1.8.0",
+    "escope": "^3.4.0",
```
hzoo: If you update babel-eslint to 4.1.8 you can remove this. https://github.com/babel/babel-eslint/issues/243

gshively11: Thanks @hzoo, I'll tackle that in my branch when I get time to start it.

PR #9, opened 2016-02-19
Move plugins to normal dependencies.

mpaulucci:
If we set the eslint plugins as normal dependencies they will be included in the `node_modules` folder of the consumer and we will be able to avoid having to specify them in the `package.json` (of the consumer, like we do now). This should work with `npm@3`, since dependencies are flattened. For `npm@2`, we still have to include them, so let's keep them as `peerDependencies` as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,13 +33,11 @@
     "babel-eslint": "^4.1.3",
     "escope": "^3.4.0",
     "eslint": "^2.1.0",
+    "eslint-plugin-mocha": "^2.0.0",
+    "eslint-plugin-react": "^3.16.1",
+    "eslint-plugin-json": "^1.2.0",
     "fashion-show": "^3.3.0",
     "install": "^0.4.4",
     "jscs": "^2.10.1"
-  },
-  "devDependencies": {
-    "eslint-plugin-mocha": "^2.0.0",
-    "eslint-plugin-react": "^3.16.1",
-    "eslint-plugin-json": "^1.2.0"
   }
 }


gshively11:
:+1:

gshively11:
we should make that change here too: https://github.com/godaddy/javascript/blob/master/package.json

sleparc:
+1

samshull:
Can you remove them from `peerDependencies` and add this to `package.json`:
```
"engine": {
  "npm": ">=3.0.0"
},
"engineStrict": true
```

gshively11:
@samshull  We're still on npm 2, I'd prefer not to force that just yet.

samshull:
So this configuration has been tested on multiple repos with both `npm@2` and `npm@3`?

samshull:
I think that keeping them as `peerDependencies` is a bit confusing for the end user of `npm@2` because they are installed, but unless the user is using dedupe to flatten their dependency tree, they still have to be aware that these are `peerDependencies` and include them separately in their `package.json`. It would be better just to remove them as `peerDependencies` and specify them as part of the install instructions for `npm@2` in the README.

mpaulucci:
I've tried testing it using `npm link` but it doesn't work there cause the dependencies of the linked module are not copied over to the `node_modules` of the consumer. I think that's because of how `npm link` works. I don't know how to test this otherwise. Anyway, I think that this PR is unlikely to break anything.

samshull:
It's not because of `npm link` its because of how `eslint` looks for/finds the plugins packages in the dependency tree. It doesn't traverse down into this packages directory looking for the plugins.

gshively11:
When you `npm install --save godaddy-style`, npm2 will automatically add the peer dependencies to your package.json as well, IIRC.

gshively11:
I think calling it out in the README is a good idea, but keeping them as peer dependencies is also useful.

mpaulucci:
+1 to keep the peer dependencies, since they are installed automatically, npm@2 users dont have to do anything special.

samshull:
I haven't tested it yet and am not a fan of "I think", can we publish this PR as a beta version in order to ensure this is all correct?

mpaulucci:
Sounds good to me.

samshull:
published as `2.0.2-beta` please test with both `npm@2` and `npm@3` that it is working as expected.

gshively11:
Test run for npm@2.  Removed existing dependency from project, did `npm install --save-dev godaddy-style@2.0.2-beta`.  My package.json was updated with the package and it's peer dependencies.  Everything worked as expected.

mpaulucci:
Tested in npm@3. Works as expected.

samshull:
+1

PR #11, opened 2016-02-19
Remove arrow-parens style enforcement.

gisenberg:
I'd like to remove the `arrow-parens` option from the style guide. Removing enforcement of a particular style allows for cases where brevity preferred:

```javascript
arr.map(i => doSomething(i));
```

... as well as providing additional clarity in these scenarios:

```javascript
const someFunction = (name) => { /* ... */ }

export default (foo) => { /* ... */ }
```

instead of the currently enforced:

```javascript
const someFunction = name => { /* ... */ }

export default foo => { /* ... */ }
```

Thoughts?

### DIFF
--- a/dist/.eslintrc
+++ b/dist/.eslintrc
@@ -17,7 +17,6 @@
   },
   "rules": {
     "accessor-pairs": 2,
-    "arrow-parens": [2,"as-needed"],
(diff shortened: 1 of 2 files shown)


indexzero:
+1

samshull:
+1

PR #13, opened 2016-03-02
[bug] Fix binPath when using the --es5 option

gshively11:
This fixes a bug that was introduced in 2.0.3, which caused a broken binPath when using the `--es5` option, due to [this code](https://github.com/indexzero/fashion-show/blob/master/lib/defaultify.js#L53) in fashion-show.  The bug only appears if you don't have `jscs` and `eslint` installed globally (which I did on my development box), or anywhere else on your existing PATH.  It would manifest like this:
```
Error: spawn jscs ENOENT
    at exports._errnoException (util.js:874:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:178:32)
    at onErrorNT (internal/child_process.js:344:16)
    at doNTCallback2 (node.js:439:9)
    at process._tickCallback (node.js:353:17)
    at Function.Module.runMain (module.js:469:11)
    at startup (node.js:134:18)
    at node.js:961:3
```

### DIFF
--- a/bin/eslint
+++ b/bin/eslint
@@ -1,12 +1,13 @@
 #!/usr/bin/env node
 
 // Remove this when es5-fix.js is no longer necessary
-var getConfigDir = require('../es5-config-dir');
+var paths = require('../es5-paths')();
 
 require('fashion-show')({
   commands: 'eslint',
-  configDir: getConfigDir(),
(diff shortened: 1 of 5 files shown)


jcrugzz:
LGTM :+1:

samshull:
+1

PR #15, opened 2016-03-04
Enhanced JSX and JSCS rules

### LINKED ISSUE #6: Consider including common eslint-plugin-react rules
From recommended rules [the package itself exports](https://github.com/yannickcr/eslint-plugin-react/blob/master/index.js#L45-L69): 

``` json
    "react/display-name": 2,
    "react/jsx-no-duplicate-props": 2,
    "react/jsx-no-undef": 2,
    "react/jsx-uses-react": 2,
    "react/no-danger": 2,
    "react/no-deprecated": 2,
    "react/no-did-mount-set-state": [2, "allow-in-func"],
    "react/no-did-update-set-state": [2, "allow-in-func"],
    "react/no-direct-mutation-state": 2,
    "react/no-is-mounted": 2,
    "react/no-unknown-property": 2,
    "react/prop-types": 2,
    "react/react-in-jsx-scope": 2,
```

samshull:
closes #6


mpaulucci:
+1

indexzero:
I am +1 here, but I believe this would make us `3.0.0`

package.json
```diff
@@ -35,13 +36,15 @@
     "eslint": "^2.1.0",
     "eslint-plugin-json": "^1.2.0",
     "eslint-plugin-mocha": "^2.0.0",
-    "eslint-plugin-react": "^3.16.1",
+    "eslint-plugin-react": "^4.1.0",
```
gingur: This should either be a dependency or a peer/dev dependency, not all three

gingur:
I agree with @indexzero, any build that is referencing this lib `^2.0.0` will all the sudden fail after this change, due to either eslint rules or dependency requirements.

package.json
samshull: @gingur This was added in #9 because of the differences between `npm@2` and `npm@3` by @mpaulucci and @gshively11. It allows us to have the desired package setup for each version of `npm`. It seems strange, but it doesn't hurt anything.

samshull: Also, it has been setup like this since 2.0.2 https://github.com/godaddy/javascript/commits/master

gingur: I've actually encountered race conditions with this setup before, I get warnings randomly when installing this package that the peer dependencies have not been met.  Having purely just peer dependencies works in both npm 2 and npm 3, this isn't an actual issue.  If using npm 3 and using this as a dependency then yes, it is expected that you must manually install the peer dependency packages; that is its intended purpose and functionality.

PR #18, opened 2016-03-09
Allow shorthand property functions on objects

### LINKED ISSUE #14: Fundamentally opposing rules for ES6
I noticed that there is a situation in ES6 where properties of an object literal are declared without the `function` keyword and because object keys are not allowed to have space afterwards, but opening round braces are not permitted to *NOT* have whitespace before, there is a situation created that causes issues with autofix and cannot be resolved manually either.

https://github.com/godaddy/javascript/blob/master/dotfiles/.jscsrc#L279
https://github.com/godaddy/javascript/blob/master/dotfiles/.jscsrc#L70

samshull:
fixes #14

### DIFF
--- a/dist/.jscsrc
+++ b/dist/.jscsrc
@@ -11,7 +11,8 @@
   "requireSpacesInConditionalExpression": true,
   "requireSpacesInAnonymousFunctionExpression": {
     "beforeOpeningRoundBrace": true,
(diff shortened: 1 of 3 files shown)


DullReferenceException:
YEEES! :guitar: :+1:

JosephJNK:
+1

PR #20, opened 2016-03-10
remove no-useless-concat rule

JosephJNK:
This rule prevents strings literals from being concatenated. It prevents you from building long strings line by line, though. Code like:
```
var x = "Say I'm formatting a really long string here, " +
          "to pass to a service that takes a textual input, " +
          "or to create a verbose error message in an odd edge case " +
          "and I don't want to make the line length absurdly long";
```
produces an error. Conforming to the rule would make that line of code into
```
var x = "Say I'm formatting a really long string here, to pass to a service that takes a textual input, or to create a verbose error message in an odd edge case and I don't want to make the line length absurdly long";
```
which is much less readable, and also violates other eslint rules (max line length). The [documentation for the rule](http://eslint.org/docs/rules/no-useless-concat) gives the justification that it prevents lines like `var x = 'a' + 'b';`, which may result accidental refactoring, but I've never seen a case like this occur in practice. I think that requiring manually disabling the rule in comments for a relatively common case of producing multiline strings is more distraction/visual noise than its worth to rule out the hypothetical case given in the rule's description.

### DIFF
--- a/dotfiles/.eslintrc
+++ b/dotfiles/.eslintrc
@@ -108,7 +108,7 @@
     "no-unused-expressions": 0,
     "no-use-before-define": [2, "nofunc"],
     "no-useless-call": 2,
-    "no-useless-concat": 2,
+    "no-useless-concat": 0,
     "no-with": 2,
     "prefer-const": 2,
     "radix": 2,


jcrugzz:
@JosephJNK this rule applies only to using the `+` for concatenation? While im pretty +0 here I'll play devils advocate and say that using template strings does solve this.

gshively11:
According to the [docs](http://eslint.org/docs/rules/no-useless-concat), multiple lines shouldn't trigger the rule.  Or am I misinterpreting the docs?  When it says "The following patterns are not considered problems:", does it mean the rule won't fire, or that the rule will fire but you can safely ignore it?

gshively11:
Because I'm pretty sure I have a project using godaddy-style with multiline strings and it doesn't complain.

JosephJNK:
@gshively11 d'oh, I missed this exemption. Closing this.

indexzero:
-1 there is a much easier way to do this that doesn't violate the rule, which prevents other badness besides this.

``` js
var x = ["Say I'm formatting a really long string here, ",
  "to pass to a service that takes a textual input, ",
  "or to create a verbose error message in an odd edge case ",
  "and I don't want to make the line length absurdly long"].join(' ')
```

JosephJNK:
@indexzero good call, I forgot about that trick.

PR #21, opened 2016-03-10
remove 'radix' rule

JosephJNK:
The "radix" rule requires the radix second argument to be passed to calls to `parseInt`.
This means that `parseInt(foo);` is invalid, but `parseInt(foo, 10);` is allowed.

This is to prevent issues in ES4 code where `parseInt` would unexpectedly parse strings with leading zeroes as octal. AFAIK, no one at GoDaddy is actively developing ES4 code though, so this rule doesn't add value.

### DIFF
--- a/dotfiles/.eslintrc
+++ b/dotfiles/.eslintrc
@@ -111,7 +111,7 @@
     "no-useless-concat": 2,
     "no-with": 2,
     "prefer-const": 2,
-    "radix": 2,
+    "radix": 0,
     "react/jsx-uses-react": 1,
     "strict": [2, "global"],
     "use-isnan": 2,


indexzero:
-1. If you are using ES6 I would suggest using `Number.parseInt` instead. https://github.com/eslint/eslint/issues/3364

JosephJNK:
@indexzero I'm using ES5 currently. Why the -1?

indexzero:
We're throwing the baby out with the bath water here. I think a change of `[2, "as-needed"]` is probably more prudent. http://eslint.org/docs/rules/radix#rule-details

JosephJNK:
@scommisso has pointed out that this could cause issues when our code runs in old browsers, even if we aren't targetting them specifically, I'll close this.

PR #23, opened 2016-03-12
Reduce `prefer-const` to a warning.

indexzero:
There are situations where ESLint cannot determine it is actually not `const`. e.g. ESLint
does not understand the semantics of `mocha`, specifically the `beforeEach`:

``` js
describe('Any test suite', () => {
  let fixture;

  beforeEach(() => {
    fixture = testHelper({ setup: 'some stuff' });
  });

  it('my first test', () => { });
  it('my second test', () => { });
});
```

The above would throw an error for `prefer-const` but fixture is actually
set twice because of how `before` works in `mocha`.

### DIFF
--- a/dist/.eslintrc
+++ b/dist/.eslintrc
@@ -107,7 +107,7 @@
     "no-useless-call": 2,
     "no-useless-concat": 2,
     "no-with": 2,
-    "prefer-const": 2,
+    "prefer-const": 1,
     "radix": 2,
     "react/jsx-uses-react": 1,
     "strict": [2,"global"],
--- a/dist/es5/.eslintrc
+++ b/dist/es5/.eslintrc
@@ -187,7 +187,7 @@
     "no-useless-call": 2,
     "no-useless-concat": 2,
     "no-with": 2,
-    "prefer-const": 2,
+    "prefer-const": 1,
     "radix": 2,
     "react/jsx-uses-react": 1,
     "strict": [
--- a/dotfiles/.eslintrc
+++ b/dotfiles/.eslintrc
@@ -110,7 +110,7 @@
     "no-useless-call": 2,
     "no-useless-concat": 2,
     "no-with": 2,
-    "prefer-const": 2,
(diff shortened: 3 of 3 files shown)


DullReferenceException:
Strange, I use this pattern constantly and have never received a warning, but I definitely receive it in other cases where I really can use `const`.

indexzero:
@DullReferenceException looks like a subtle bug in ESLint. Opened a bug for it: https://github.com/eslint/eslint/issues/5555. You can also read [the reproduction of the issue](https://gist.github.com/indexzero/0df4264d5a1121469d2e)

apoco:
Very interesting.  Am I understanding correctly that this will only happen for variables not actually read outside of the assignment scope? If you refer to the variable in one of the it blocks, I assume the warning would not happen.

It is a mislabeled error, but the sample code is bad because fixture is not even needed in the closure. If it only occurs in those cases, I'd say it's not worth downgrading the rule.

indexzero:
Yes. That's correct. The discussion I had with the eslint team indicated the behavior for this rule is going to change as follows:

> Never prefer-const for uninitialised let statements

This makes sense since uninitialised const statements are invalid JavaScript. (See: https://github.com/eslint/eslint/issues/5284). 

As such I'm going to merge this until that bug is fixed unless there are any strong objections.

apoco:
I'd personally prefer it was an error still since there is no legitimate reason for the sample code as given. If the it block referenced the variable and eslint still complained, I'd fully agree. But the message is incorrect, so I guess +0 on balance.

indexzero:
Again once the bug is fixed I'm fine changing this back to an error. Until then unassigned let statements move over into the land of style and not semantic. I've seen this from some other folks as well. I'll let them follow up with +1s

3rd-Eden:
+1

jcrugzz:
Yea this feels too restrictive and prevents using `let` for valid use cases. :+1:

PR #25, opened 2016-03-17
Add instructions to override a specific rule locally

gzurbach:
We have been using GoDaddy's style guide in 4 projects so far and we always encounter the need to override one or two eslint rules.

We struggled quite a bit to understand how to extend and override the provided `.eslintrc` file in `dist`. We thought it would be nice to share our findings with everyone else.

-- The MYA and API teams

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ This section is a place for additional best practices that may be useful but are
 ### How do I override a specific rule ?
 
 1. Add a `.eslintrc` file at the root of your project:
-```
+``` js
 {
   "extends": "./node_modules/godaddy-style/dist/.eslintrc",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,6 +73,28 @@ This section is a place for additional best practices that may be useful but are
 
 ## FAQ
 
+### How do I override a specific rule ?
+
+1. Add a `.eslintrc` file at the root of your project:
+```
```
3rd-Eden: Should also use `js` as language here as this is JSON.

gzurbach: @3rd-Eden thanks, done!

Swaagie:
+1

Issue #27, opened 2016-03-31
Disable JSCS auto-fix?


gzurbach:
We have the following npm script in our `package.json`:

``` json
"scripts": {
  "lint_local": "godaddy-js-style-eslint -c .eslintrc server common test && godaddy-js-style-jscs server common test",
  "lint_jenkins": "godaddy-js-style-eslint -c .eslintrc -f checkstyle server common test -o checkstyle-eslint.xml && godaddy-js-style-jscs -r checkstyle server common test > checkstyle-jscs.xml"
}
```

In Jenkins, we run `npm run lint_jenkins`. We expect that `checkstyle-eslint.xml` and `checkstyle-jscs.xml` will be reporting style issues. But instead because [`--fix` is set by default](https://github.com/godaddy/javascript/blob/master/bin/jscs#L11) by `godaddy-js-style-jscs`, our source files get fixed during the build. This is not the desired behavior. 

Would you guys accept a PR adding an option to disable auto-fix? I'm not sure how it would play with `fashion-show` (might be tricky). The change would be backward-compatible, and would allow us to stop our build if the style guide is not respected.

indexzero:
Making this an option is definitely a feature that makes sense. The [relevant code is here](https://github.com/godaddy/javascript/blob/master/bin/jscs#L11). 

Something to consider w.r.t. your overall process: if you do this:

``` js
"scripts": {
  "pretest": "npm run lint_local",
  "test": "echo 'whatever your test script is'",
  "lint_local": "godaddy-js-style-eslint -c .eslintrc server common test && godaddy-js-style-jscs server common test",
  "lint_jenkins": "godaddy-js-style-eslint -c .eslintrc -f checkstyle server common test -o checkstyle-eslint.xml && godaddy-js-style-jscs -r checkstyle server common test > checkstyle-jscs.xml"
}
```

Then code will be auto-fixed before committing and getting to Jenkins.

indexzero:
Fixed in #40.

PR #28, opened 2016-03-31
Add option to disable JSCS auto-fix

gzurbach:
See https://github.com/godaddy/javascript/issues/27 for details.


Swaagie:
Not a show stopper but as for argument parsing wouldn't it be simpler to use an argument parsing module? Although its just a single argument atm, more might follow in the future. That said this works, so +1, would like someone else to confirm as well.

jcrugzz:
I agree, at this point we should use something like [`yargs`](https://www.npmjs.com/package/yargs) or [`minimist`](https://www.npmjs.com/package/minimist) to parse the arguments for both the `es5` arg and for this so we are only doing parsing in one place.

indexzero:
@gzurbach could you use `minimist` instead? https://github.com/substack/minimist ... it has zero dependencies and sets us up to be a bit more future proof for future options we want to include. It would also remove the need for a separate file to parse options.

gzurbach:
No problem. I will update my PR soon.

jscs/auto-fix.js
```diff
@@ -0,0 +1,14 @@
...
+
+module.exports = jscsAutoFix;
+
+function jscsAutoFix() {
+  var args = process.argv.slice(2);
+  var noFixArgIndex = args.indexOf('--noFix');
+  var noFix = noFixArgIndex > -1;
+
+  if (noFix) {
+    process.argv.splice(noFixArgIndex + 2, 1);
```
gzurbach: Using minimist doesn't cover the need to remove the `--noFix` param before passing it to `fashion-show`. Should I keep this file around to maintain that logic?

jcrugzz: we should handle automatically stripping arguments that minimist expects and pass that to `fashion-show` rather than modifying `process.argv`. That can be handled before or after the minimist bit.

bin/lint
```diff
@@ -3,12 +3,14 @@
 // Remove this when es5/fix.js is no longer necessary
 var paths = require('../es5/paths')();
 
+var autoFix = require('../jscs/auto-fix')();
+
 require('fashion-show')({
   commands: ['jscs', 'eslint'],
   configDir: paths.configDir,
   targets: process.argv.splice(2),
```
indexzero: This would simply become `argv._`

jscs/auto-fix.js
indexzero: @gzurbach once an options parser is introduced to this data flow we can take advantage of it in useful ways to avoid the need for modifying `process.argv` at all. e.g. [my other comment](https://github.com/godaddy/javascript/pull/28/files/12a8982f29f098d4ab581eda49aaaf16a07dd0fc#r58423483)

``` js 
 targets: process.argv.splice(2),
```

simply becomes

``` js
  targets: argv._
```

gzurbach:
I updated my PR but `argv._` is not gonna work for me. My lint scripts look like this:

```
godaddy-js-style-eslint -c .eslintrc server common test
godaddy-js-style-jscs -r checkstyle server common test
```

Using `argv._` strips out `-c .eslintrc` and `-r checkstyle` which allows me to pass my own eslint config file (which extends the GD config) or change the output format for Jenkins.

indexzero:
Interesting. I would file those use cases under "worked by accident." Let me think about how to support those and this at the same time. Will get back to you tomorrow.

gzurbach:
@indexzero did you get a chance to look at this?

indexzero:
@gzurbach unfortunately I have not had a chance yet to really think through it yet. My current thought process is to expose programmatically in `fashion-show` a way to optionally specify either the `jscs` or `eslint` config and then pass those down in our binary scripts.

indexzero:
This was fixed in #40 when `jscs` was removed.

PR #29, opened 2016-04-01
Update dependencies

Swaagie:
Update eslint and babel-eslint to fix the estraverse-fb dependency issue.
To pass our own styleguide/tests the `'use strict'` was required.

This works for ES6 would like someone that still uses ES5 to confirm.

Also had to change `"sourceType": "string"` to `"sourceType": "module"` for the linting to work on es6 import styles at all. I think this is fine since we create mostly modular es6 modules anyways that are then later packed/bundled with builders.


jcrugzz:
This LGTM :+1:. Still passes on an es5 project

indexzero:
+1

PR #30, opened 2016-04-01
Disable `react/display-name` for ES6

mpaulucci:
Since the recommended way of creating Components is using ES6 classes.


DullReferenceException:
:+1:

indexzero:
+1 ... all ES6 everything these days. The `.prototype` is dead! Long live the `.prototype`!

PR #31, opened 2016-05-26
[deps] Remove useless peerDependencies, bump other dependencies to la…

3rd-Eden:
Makes sure we're up to date with all latest packages and removes pointless peerDependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,23 +24,17 @@
     "jscs",
     "eslint"
   ],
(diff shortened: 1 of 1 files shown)


indexzero:
If you would please resubmit against the latest `master` this seems like the correct approach. The current incantation was modeled after `airbnb/javascript` which takes advantages of `peerDependencies`.

PR #36, opened 2016-12-29
turn ”react/prefer-es6-class" on as an error

yhan64:
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-es6-class.md


mpaulucci:
👍

es5/fix.js
```diff
@@ -29,7 +29,8 @@ var es5Options = {
       sourceType: 'script'
     },
     rules: {
-      'react/display-name': 2
+      'react/display-name': 2,
+      "react/prefer-es6-class": 0,
```
yhan64: @mpaulucci Does this work?

mpaulucci: it should! 👍

indexzero:

PR #38, opened 2016-12-29
turn on "react/jsx-pascal-case" as an error

yhan64:
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-pascal-case.md


mpaulucci:
👍

indexzero:

PR #39, opened 2017-01-27
Lessen the `no-invalid-this` to a warning

indexzero:
There are a few motivations here:

1.  `this` is valid when binding to class functions in ES6.
2. `this` inside of `describe` for `mocha` (and other test frameworks) is valid.


rkmarks:
👍

indexzero:

Issue #32, opened 2016-06-14
JSCS end of support


gzurbach:
Hi all,

The JSCS team announced a couple months ago that they are not maintaining their project anymore (see https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2):

> We came to the realization that ESLint and JSCS are really solving the same set of problems in very similar ways. The community seems to want to use a single linter, and we want to help that become a reality.

Therefore I am wondering if it still makes sense to have an abstraction layer for those two linting tools?

As an API developer writing code in Node only, I ran into several issues using this package:

- It is not possible to override specific rules (in fact it works but it's a coincidence - see #25): one size does not fit all! Some frameworks such as Loopback impose us to have methods with 5+ params all over, but the GoDaddy ESLint complains.
- It was made by and for React developers: several linting rules and eslint plugins included in this project are only useful for React/JSX folks. This default configuration makes it difficult for Node users to use this project.
- It cannot be used on Jenkins: ESLint options to change the output format or to configure the output file cannot be passed via `fashion-show` (see #28).

Since JSCS is now deprecated, I wonder if this project should simply contain the following:

1. A base ESLint configuration file for all GoDaddy teams
2. A base ESLint configuration file for React developers

That way the entire company could use it and align on a base style. Each team would install ESLint directly in their project (locally or globally). Each project would extend the base ESLint configuration files such as:
```js
{
  "extends": "./node_modules/godaddy-js-style-guide/.eslintrc-base",
  "globals": {
    "myProjectGlobal": true
  },
  "rules": {
    "max-params": 0, // because I <3 params!
  }
}
```

Just my $0.02, cheers! :)

indexzero:
I am -1 on this until there is parity with `eslint` and `jscs` for autofixing. This is something the `jscs` team is working on now that they have joined forces with the `eslint` folks. Going to leave this open as a way for us to track their progress.

indexzero:
Good news! The parity is here: https://gist.github.com/indexzero/a51afcf01fda7fd2af81932fe72b833c

Issue #17, opened 2016-03-08
Does not run with eslint 2.3.0


gzurbach:
A bug was introduced in the latest version of `eslint` (2.3.0), preventing us from running the style guide.

```
Running eslint with options: -c godaddy-style/dist/.eslintrc server common test
Error: Cannot find module 'estraverse-fb'
```

See issue: https://github.com/eslint/eslint/issues/5476

I'm sure this issue will be fixed in a couple days, but I wonder whether it would be wise to lock the dependencies in the `package.json` (i.e. change `"eslint": "^2.1.0"` to `"eslint": "2.2.0"`). 

I tend to run into less issues when controlling which versions I'm using. Just a suggestion :)

gshively11:
Hi @gzurbach.  Based on that eslint issue, it definitely seems like something is amiss, however, I'm unable to reproduce it locally, on either v2 or v3 of godaddy-style.  Can you provide any more details about this one?  Your package.json would be a good start.  I might have a dependency that is masking the problem.

gzurbach:
@gshively11 You likely have eslint 2.1 or 2.2 in your local npm cache. Since the guide specifies `"eslint": "^2.1.0"` your cache has a match and doesn't get the latest version. Try to run `rm -rf node_modules && npm cache clear && npm i`, you should then be able to see the bug.

hzoo:
babel-eslint isn't compatible with ESLint 2.3.x at the moment (I pushed a change that removes the specific error, but it still might be broken) so I would recommend pinning.

gshively11:
@gzurbach I believe I did that, but I don't recall for sure.  I'm a bit swamped at the moment, will investigate this within a few days.  Apologies for the delay.

indexzero:
Fixed in #40.

Issue #16, opened 2016-03-08
Running any of the scripts fails in event.js:85


cwillwer:
I have attempted to follow the directions for adding it to our project, but running any of the commands with `npm run <script>` results in

```
events.js:85
      throw er; // Unhandled 'error' event
            ^
```

I assume that events.js is from one of the node_modules that this depends on because we do not have an events.js file in our code. I changed the `lib/` to `src/` in the script command because we don't have a lib folder, and got the same behavior. npm error information looks like this:

```
npm ERR! Windows_NT 6.1.7601
npm ERR! argv "c:\\Program Files\\nodejs\\node.exe" "c:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run" "lint"
npm ERR! node v0.12.7
npm ERR! npm  v2.11.3
npm ERR! code ELIFECYCLE
npm ERR! <our package name>@1.0.0 lint: `godaddy-js-style-lint src/ test/`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the <our package name>@1.0.0 lint script 'godaddy-js-style-lint src/ test/'.
npm ERR! This is most likely a problem with the <our package name> package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     godaddy-js-style-lint src/ test/
```

gzurbach:
I think the `lib/` was just an example, you have to adjust it to your own use case.

There should be an error above, a call stack maybe. Could you post it?

cwillwer:
Yes, that's why I said that I replaced `lib/` with `src/` because we have a `src` and `test` subdirectory parallel with our package.json file.

The callstack is:
```
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: spawn jscs ENOENT
    at exports._errnoException (util.js:746:11)
    at Process.ChildProcess._handle.onexit (child_process.js:1053:32)
    at child_process.js:1144:20
    at process._tickCallback (node.js:355:11)
    at Function.Module.runMain (module.js:503:11)
    at startup (node.js:129:16)
    at node.js:814:3
```

BTW, for people who are not knowledgeable in running npm scripts, it would be helpful if the documentation mentioned the `npm run <cmd>` command to run the script entries in the package.json file.

gzurbach:
I've actually never used the `godaddy-js-style-lint`. Have you tried with `godaddy-js-style-eslint`?

gshively11:
[...]

PR #40, opened 2017-02-01
ESLint everywhere

indexzero:
> ⚠️ ✋ Work in progress, but ready for an initial review ✋ ⚠️ 

> Fixes all open issues: #16, #17, #27, #32

This PR removes `jscs` and replaces all rules with `eslint` auto-fix. See [`jscs` to `eslint` research](https://gist.github.com/indexzero/a51afcf01fda7fd2af81932fe72b833c). Since this now uses a single platform, `fashion-show` is no longer necessary. 

The new API surface area is three `eslint-config-**` packages:

- `eslint-config-godaddy`: Base configuration for _non-React_, ES6 JavaScript applications
- `eslint-config-godaddy-react`: Configuration for ES6 React JavaScript applications
- `eslint-config-godaddy-es5`: Configuration for React _and_ non-React ES5 JavaScript applications.


package.json
```diff
@@ -27,24 +28,19 @@
   "peerDependencies": {
     "eslint-plugin-mocha": "^2.0.0",
     "eslint-plugin-react": "^4.1.0",
-    "eslint-plugin-json": "^1.2.0",
```
3rd-Eden: These files should become dependencies instead of peer or devDependencies as our configuration extend these files. See https://github.com/godaddy/javascript/pull/31

samshull:
👏 Thanks for the work @indexzero! Looks great!

rkmarks:
👍 great work!

chemdrew:
👍

package.json
indexzero: I was following the pattern established in [`airbnb/javascript` for `peerDependencies`]. 

_... puts on thinking cap ..._ 

If we want to be backwards compatible for both `npm@2` *AND* `npm@3` then I think we need to keep these in `peerDependencies` (but obviously make them more liberal – like `"*"`). If we want to _only_ support `npm@3` then we can keep them as `dependencies` since they will be flattened by default.

Another option (which I have NO idea will work) is to use the `require.resolve` pattern used for `"extends"` for `"plugins"`. i.e. change:

``` js
{
  plugins: ['mocha', 'json'],
}
```

to use `require.resolve`:

``` js
{
  plugins: ['eslint-plugin-mocha', 'eslint-plugin-json']
    .map(require.resolve),
}
```

[`airbnb/javascript` for `peerDependencies`]: https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb#eslint-config-airbnb-1

rkmarks: As a side note I am definitely all for maintaining backwards compatibility. Option 2 is definitely an unknown.

Issue #42, opened 2017-03-09
4.0.0 fails to install


3rd-Eden:
```
npm ERR! node v7.5.0
npm ERR! npm  v4.1.2
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn

npm ERR! godaddy-style@4.0.0 postinstall: `run-p install:config install:config:es5 install:config:react`
npm ERR! spawn ENOENT
npm ERR!
npm ERR! Failed at the godaddy-style@4.0.0 postinstall script 'run-p install:config install:config:es5 install:config:react'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the godaddy-style package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     run-p install:config install:config:es5 install:config:react
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs godaddy-style
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls godaddy-style
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/V1/Projects/ux/svgs/npm-debug.log
```

The `run-p` command does not exist on my system so it fails to install .

indexzero:
Thanks, but just FYI 4.0.0 of this package is deprecated. Use the sub packages it exposes:

- eslint-config-godaddy
- eslint-config-godaddy-react
- eslint-config-godaddy-es5

3rd-Eden:
I know, but without being able to install the package, people wont be able to see the deprecation notice. A Fix was published as 4.0.1 so i'm closing this.

PR #43, opened 2017-03-09
[fix] Include npm-run-all as dependency

3rd-Eden:
Fix for #42. The `postinstall` hook of the package uses a `run-p` binary to run various of tasks after the installation of the package. `run-p` is a cli item that is shipped in the `npm-run-all` package which is currently installed as devDependency instead of normal dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,9 @@
     "eslint-find-rules": "^1.14.3",
     "eslint-plugin-json": "^1.2.0",
     "eslint-plugin-mocha": "^4.8.0",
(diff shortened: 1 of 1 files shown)


jcrugzz:
LGTM 👍

indexzero:
Thanks, but just FYI 4.0.0 of this package is deprecated. Use the sub packages it exposes:

- eslint-config-godaddy
- eslint-config-godaddy-react
- eslint-config-godaddy-es5

PR #45, opened 2017-04-21
[minor] enable object rest spread

Swaagie:
ES7, but we all love it right? yay 👍 or nay 👎


bmeck:
Relevant: https://groups.google.com/forum/#!topic/v8-users/ricllIFdP2w v8/chrome intent to ship

PR #46, opened 2017-05-22
Indenting case clauses two spaces

johnobe:
This PR alters eslint's `indent` rule to force `case` clauses within switch statements to be indented two spaces.

For example

```js
switch (action.type) {
case MY_ACTION:
  ...
  break;
default:
  ...
}
```

will now need to be written as

```js
switch (action.type) {
  case MY_ACTION:
    ...
    break;
  default:
    ...
}
```

For more information on `indent`'s options see http://eslint.org/docs/rules/indent


johnobe:
@indexzero, @Swaagie - any opposition to this being merged? If you're waiting on me to merge, my account is not in the org.

PR #48, opened 2017-07-20
Rule change suggestion for 'object-curly-spacing'

fabiomcosta:
Code with `objectsInObjects: false`, looks wrong with object destructuring and a default object value:

Here is an example:

Code with `objectsInObjects: false`:
```js
const { org = {}} = config;
```

Without:
```js
const { org = {} } = config;
```

### DIFF
--- a/packages/eslint-config-godaddy/index.js
+++ b/packages/eslint-config-godaddy/index.js
@@ -139,9 +139,7 @@ module.exports = {
     'indent': [2, 2, { SwitchCase: 1 }],
     'keyword-spacing': 2,
     'linebreak-style': 2,
-    'object-curly-spacing': [2, 'always', {
-      objectsInObjects: false
-    }],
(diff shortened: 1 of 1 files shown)


indexzero:
@fabiomcosta can you show examples with objects, not just destructuring assignment?

yungcheng:
a related question, I actually want this rule to be `never` 
```json
'object-curly-spacing': [2, 'never']
```

https://eslint.org/docs/rules/object-curly-spacing#never

I can send a PR for it. However, before I do that, I want to know if there are previous discussions around this.

rkmarks:
@indexzero I believe the same happens when using an object like in `style={{}}` that's just one example but would look better with additional spacing of the object.

fabiomcosta:
Thx folks, unfortunately I don't have rights to merge though.

rkmarks:
@indexzero any reason why we cannot merge this change?

PR #49, opened 2017-07-20
Update `array-bracket-spacing` rule

fabiomcosta:
Given we have `object-curly-spacing`  as 'always', shouldn't `array-bracket-spacing` also be 'always' for consistency?

Examples on ESlint's page: http://eslint.org/docs/rules/array-bracket-spacing#always

My personal opinion on this honestly is to do the opposite, I would set `object-curly-spacing: 2` (defaults to 'never').
One way or the other, I think these 2 should be consistent.

### DIFF
--- a/packages/eslint-config-godaddy/index.js
+++ b/packages/eslint-config-godaddy/index.js
@@ -125,7 +125,7 @@ module.exports = {
     // Whitespace rules follow. All rules are auto-fix unless
     // otherwise specified.
     //
-    'array-bracket-spacing': 2,
+    'array-bracket-spacing': [2, 'always'],
     'brace-style': [2, '1tbs', { allowSingleLine: true }],
     'camelcase': [2, { properties: 'never' }],
     'comma-spacing': 2,


fabiomcosta:
I'll gladly fix the build errors when I get some feedback on this.

dudeinthemirror:
I personally have the same preference: no spaces (for both cases). But I agree that it should be consistent.

indexzero:
Being motivated by consistency makes sense to me, but these are different JavaScript semantics @fabiomcosta so the consistency argument does not make sense.

fabiomcosta:
@indexzero so you're saying this looks consistent?

```js
// object-curly-spacing: always
const { a } = { a: 1, b: 2, c: 3 };

// array-bracket-spacing: never
const [a] = [1, 2, 3];
```

indexzero:
Indeed. They are consistent in the way that these are consistent

``` js
function foo() { return 'bar'; }

if (conditional || somethingElse) {
  console.log('if statements and functions');
} else {
  console.log('are different language semantics');
  console.log('so their style need not be the same');
}
```

PR #53, opened 2017-08-07
Prevent hard version dependencies between packages

samshull:


fabiomcosta:
`eslint-config-godaddy-es5` will also have to be updated

PR #56, opened 2017-12-11
feat(arrow-fn): require parens

gergelyke:
To enforce consistency, requires parens for every arrow functions

### DIFF
--- a/packages/eslint-config-godaddy/index.js
+++ b/packages/eslint-config-godaddy/index.js
@@ -20,6 +20,7 @@ module.exports = {
   rules: {
     'accessor-pairs': 2,
     'arrow-spacing': [2, { before: true, after: true }],
+    'arrow-parens': [2, 'always'],
     'block-scoped-var': 2,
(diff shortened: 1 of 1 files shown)


3rd-Eden:
Seems like a massive edge check for when people accidentally `=>` instead of `>=` in if statements. I personally not +1 or -1 on this issue as I simply don't run to that case.

packages/eslint-config-godaddy/index.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   rules: {
     'accessor-pairs': 2,
     'arrow-spacing': [2, { before: true, after: true }],
+    'arrow-parens': [2, 'always'],
```
gingur: maybe we could start out with `as-needed`? do we know if autofix handles this?

JosephJNK:
-1. How does leaving parens off of arrow functions decrease code quality? I don't see any benefit here.

gergelyke:
oops, i might have jumped the gun on this.  closing for now and maybe i'll bring it up again later

PR #57, opened 2017-12-11
Add minimal styleguide

JosephJNK:
This adds a minimal styleguide which only enforces rules that teams generally agree upon. The disabled rules are those which are commonly disabled in existing internal projects, or which developers have found add friction to their development experience. The general guiding principle of this style guide is that eslint should not be configured to maximize the issues that it might detect, but rather to maximize its accuracy. That is to say, it is better for eslint to not complain about something which might be a problem than for it to complain about things which are not problems. At the end of the day, all code should go through human code review, and this is the step at which quality problems that reflect a larger design issue should be discussed.

Here is the justification behind the disabled rules:

`complexity`: This rule is disabled in some internal projects. Complexity may be artificially inflated by adding checks to verify inputs and bail early if verifications fail. Further, a function is too complex if it is hard for a human to understand, regardless of its computed cyclomatic complexity. This is an issue which should be addressed during code review.

`consistent-return`: This is an attempt at type-checking at the syntactic level, which is too low-level for the problem at hand. Further, teams who are using Flow have a better way of catching these errors already, and are intending to return a union or optional type in the cases which would trigger this rule.

`id-length`: Names should be as long as they need to be to be clear. If a name is unclear, that's an issue for code review.

`indent`: These rules are changed, not disabled, in order to give developers more freedom in how they indent certain expressions. They allow nested ternary expressions to be indented in a tabular format and for multiple `const` variables declared in a single statement to be lined up.

`max-params`: There are two cases in which this rule is violated. The first case is those rare circumstances in which a function actually requires a large number of parameters. These situations are more likely to occur when a number of small functions are generated by binding selected arguments to a large function, which is a pattern that should not be discouraged. Because there are legitimate use-cases for this, false-positives are common. The other case where argument lists are long is when interfaces are poorly designed, and additional responsibilities start to creep into the same function. In these cases, this rule is generally satisfied by stuffing additional arguments into an `options` object, complicating the interface of the function, making the function's interface more opaque, and failing to solve the underlying problem. These are problems that should be detected and discussed in code review, not mechanically.

`max-nested-callbacks`: This rule is frequently disabled because it often gives false-positives in unit tests and in code making heavy use of higher-order functions. Because this rule tries to detect a high-level concern (nested asynchronous control-flow using callbacks) using a low-level technique (syntactic analysis), false positives are common and inevitable. And cases where control flow is overly complex can easily be seen and discussed in code review.

`max-statements`: Short functions are better than long ones, but where the "too many" line is depends entirely on the function and statements in question. For some cases, 5 statements is too many, while for others, 20 is fine. The size and complexity of functions should be one of the main focuses of code review, and an automated tool can't replace this. (but can incentivize developers to prematurely cut their functions apart along bad interfaces, complicating the code review and refactoring process later.)

`no-else-return`: This is aesthetically displeasing to some, but is not a source of errors.

`no-nested-ternary`: This is a common and useful pattern. It serves the purpose of pattern matching with guards in functional languages, and helps to reduce statefulness and nested control flow by moving logic from statements to expressions. In effect it gives the benefits which `switch` statements provide over long `if`/`else` chains, without the downside of fall-through, and with the added benefit of staying in expression-land.

`no-shadow`: This is common and useful when working with higher-order functions in which an inner function works with the same type of data that exists in an outer scope. It is also common when writing tail-recursive functions, in which a binary or ternary function is needed to track the algorithm's state, but a unary function should be exposed to the caller. In these cases shadowing prevents accidental capture of the outer variable (which often leads to hard-to-debug problems) and keeps inner names from becoming long and confusing. In contrast, accidental shadowing is generally rare, and is a symptom of overly-large scopes or bad names. (issues which should be addressed in code review.) Cases of accidental shadowing can also often be caught by Flow.

`no-undefined`: This rule is pointless for codebases targeting newer platforms, because these platforms prevent `undefined` from being reassigned. We also already have rules preventing keywords from being shadowed by other variables. `foo === undefined` is more clear and expressive than `typeof foo === 'undefined'`, and there's no downside to doing the first in most contexts.

`no-unused-vars`: The `ignoreRestSiblings` argument enables a specific use-case, in which a variable is specified specifically to name which fields are unused in an object spread. Ignoring unused arguments forces developers to leave the tail of argument lists off when they only intend on using earlier arguments in the list. This reduces the clarity and expressivity of code when it is read later; it is no longer clear what the arity of the involved functions is. It is better to either give names to the unused arguments, or to mark them with a name like `_` or `__`, indicating that an argument is there, but is being intentionally unused. (as opposed to forgotten)

`spaced-comment`: This does not have any effect on code quality, and is just an aesthetic preference. It also causes a great deal of friction for many devs during the development cycle, when commenting out a piece of code or adding comments to code they're debugging causes linting to fail without the tests running. This has a concrete, negative effect on productivity.

### DIFF
--- a/packages/eslint-config-godaddy-minimal/index.js
+++ b/packages/eslint-config-godaddy-minimal/index.js
@@ -1,0 +1,24 @@
+module.exports = {
+  extends: [
+    require('./extends')('eslint-config-godaddy-react-flow')
+  ],
+  rules: {
+    'complexity': 0,
+    'consistent-return': 0,
+    'id-length': 0,
+    'indent': [2, 2, {
+      VariableDeclarator: { var: 2, let: 2, const: 3 },
+      flatTernaryExpressions: true,
+      MemberExpression: 'off'
+    }],
+    'max-params': 0,
+    'max-nested-callbacks': 0,
+    'max-statements': 0,
+    'no-else-return': 0,
+    'no-nested-ternary': 0,
+    'no-shadow': 0,
+    'no-undefined': 0,
+    'no-unused-vars': ['error', { args: 'none', ignoreRestSiblings: true }],
+    'spaced-comment': 0
+  }
+};
--- a/package.json
+++ b/package.json
@@ -4,14 +4,16 @@
   "description": "The approach to JavaScript at GoDaddy. We think it's pretty decent.",
   "scripts": {
     "lint": "eslint -c .eslintrc.js packages/**/*.js",
-    "postinstall": "run-p install:config install:config:es5 install:config:react install:config:react-flow",
+    "postinstall": "run-p install:config install:config:es5 install:config:react install:config:react-flow install:config:minimal",
     "install:config": "cd packages/eslint-config-godaddy && npm install",
     "install:config:es5": "cd packages/eslint-config-godaddy-es5 && npm install",
+    "install:config:minimal": "cd packages/eslint-config-godaddy-minimal && npm install",
     "install:config:react": "cd packages/eslint-config-godaddy-react && npm install",
     "install:config:react-flow": "cd packages/eslint-config-godaddy-react-flow && npm install",
-    "test": "run-p lint test:config test:config:es5 test:config:react test:config:react-flow",
+    "test": "run-p lint test:config test:config:es5 test:config:react test:config:react-flow test:config:minimal",
     "test:config": "cd packages/eslint-config-godaddy && npm test",
     "test:config:es5": "cd packages/eslint-config-godaddy-es5 && npm test",
+    "test:config:minimal": "cd packages/eslint-config-godaddy-minimal && npm test",
     "test:config:react": "cd packages/eslint-config-godaddy-react && npm test",
     "test:config:react-flow": "cd packages/eslint-config-godaddy-react-flow && npm test"
   },
--- a/packages/eslint-config-godaddy-minimal/.eslintrc
+++ b/packages/eslint-config-godaddy-minimal/.eslintrc
@@ -1,0 +1,3 @@
+{
+  "extends": "./index.js"
+}
--- a/packages/eslint-config-godaddy-minimal/README.md
+++ b/packages/eslint-config-godaddy-minimal/README.md
@@ -1,0 +1,55 @@
+# eslint-config-godaddy-minimal
+
+Configuration for ES6 React JavaScript applications using Flow, officially used at GoDaddy. This styleguide variant is meant to enforce a minimal set of good practices and avoid getting in the way during most workflows. There are many useful features:
+
+- **Standard. No configuration.** – Stop worrying about style and focus on your work.
+- **Modern** – Uses modern linting tools like `eslint`.
+- **Auto-fix** – Auto-fix is enabled by-default through in `eslint`. Many rules will fix themselves!
+
+This styleguide is used by dozens of product teams at GoDaddy. Have a question or comment? [Open an issue!](https://github.com/godaddy/javascript/issues/new)
+
+## Installation
+
+``` sh
+# ES6 (including React rules)
+npm i eslint-config-godaddy-minimal --save-dev
+```
+
+## Usage
+
+There are two ways to use this styleguide depending on your own tooling preference: directly using pre-included binaries or running `eslint` yourself with a custom `.eslintrc` config.
+
+##### 1. Use the pre-included binaries.
+
+These use _exactly_ the configuration defined in this  package (`eslint-config-godaddy-minimal`) **with auto-fix enabled automatically.**
+
+``` js
+{
+  "scripts": {
+    "lint": "eslint-godaddy-minimal files/ you/ want-to/ lint/"
+  }
+}
+```
+
+##### 2. Define your local `.eslintrc` and run `eslint` yourself:
+
+``` js
+module.exports = {
+  extends: 'godaddy-react-minimal',
+  rules: {
+    //
+    // Put any rules you wish to override here.
+    //
+  }
+}
+```
+
+The `--fix` option in `eslint` is [**only** available as a CLI option](https://github.com/eslint/eslint/issues/8041). Auto-fix will *NOT be enabled* unless you run `eslint --fix` in your `package.json`.
+
+``` js
+{
+  "scripts": {
+    "lint": "eslint --fix files/ you/ want-to/ lint/"
+  }
+}
+```
--- a/packages/eslint-config-godaddy-minimal/bin/eslint-godaddy-minimal
+++ b/packages/eslint-config-godaddy-minimal/bin/eslint-godaddy-minimal
@@ -1,0 +1,4 @@
+#!/usr/bin/env node
+require('eslint-config-godaddy/cli')(
+  require('path').join(__dirname, '..', 'index')
+);
--- a/packages/eslint-config-godaddy-minimal/extends.js
+++ b/packages/eslint-config-godaddy-minimal/extends.js
@@ -1,0 +1,17 @@
+const path = require('path');
+
+/**
+ * Helper function that attempts to resolve the specify named `config`
+ * locally before attempting to resolve it from `node_modules`. This
+ * tricks TravisCI into resolving correctly.
+ *
+ * @param {string} config Named `eslint-config-*` module
+ * @returns {string} Full path resolved locally or from `node_modules`.
+ */
+module.exports = function resolveLocal(config) {
+  try {
+    return require.resolve(path.resolve(__dirname, '..', config));
+  } catch (ex) {
+    return require.resolve(config);
+  }
+};
--- a/packages/eslint-config-godaddy-minimal/package-lock.json
+++ b/packages/eslint-config-godaddy-minimal/package-lock.json
@@ -1,0 +1,2230 @@
+{
+  "name": "eslint-config-godaddy-minimal",
+  "version": "2.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "acorn": {
+      "version": "5.2.1",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/acorn/-/acorn-5.2.1.tgz",
+      "integrity": "sha1-MXrHghgmwixwLWYYmrg1lnXxNdc=",
+      "dev": true
+    },
+    "acorn-jsx": {
+      "version": "3.0.1",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
+      "integrity": "sha1-r9+UiPsezvyDSPb7IvRk4ypYs2s=",
+      "dev": true,
+      "requires": {
+        "acorn": "3.3.0"
+      },
+      "dependencies": {
+        "acorn": {
+          "version": "3.3.0",
+          "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/acorn/-/acorn-3.3.0.tgz",
+          "integrity": "sha1-ReN/s56No/JbruP/U2niu18iAXo=",
+          "dev": true
+        }
+      }
+    },
+    "ajv": {
+      "version": "5.5.1",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/ajv/-/ajv-5.5.1.tgz",
+      "integrity": "sha1-s4u4h22ehr7plJVqBOch6IskjrI=",
+      "dev": true,
+      "requires": {
+        "co": "4.6.0",
+        "fast-deep-equal": "1.0.0",
+        "fast-json-stable-stringify": "2.0.0",
+        "json-schema-traverse": "0.3.1"
+      }
+    },
+    "ajv-keywords": {
+      "version": "2.1.1",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/ajv-keywords/-/ajv-keywords-2.1.1.tgz",
+      "integrity": "sha1-YXmX/F9gV2iUxDX5QNgZ4TW4B2I=",
+      "dev": true
+    },
+    "ansi-escapes": {
+      "version": "3.0.0",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/ansi-escapes/-/ansi-escapes-3.0.0.tgz",
+      "integrity": "sha1-7D6LTp+AZPwCw6ybZfHCdb2o75I=",
+      "dev": true
+    },
+    "ansi-regex": {
+      "version": "2.1.1",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/ansi-regex/-/ansi-regex-2.1.1.tgz",
+      "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
+      "dev": true
+    },
+    "ansi-styles": {
+      "version": "2.2.1",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/ansi-styles/-/ansi-styles-2.2.1.tgz",
+      "integrity": "sha1-tDLdM1i2NM914eRmQ2gkBTPB3b4=",
+      "dev": true
+    },
+    "argparse": {
+      "version": "1.0.9",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/argparse/-/argparse-1.0.9.tgz",
+      "integrity": "sha1-c9g7wmP4bpf4zE9rrhsOkKfSLIY=",
+      "dev": true,
+      "requires": {
+        "sprintf-js": "1.0.3"
+      }
+    },
+    "array-includes": {
+      "version": "3.0.3",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/array-includes/-/array-includes-3.0.3.tgz",
+      "integrity": "sha1-GEtI9i2S10UrsxsyMWXH+L0CJm0=",
+      "dev": true,
+      "requires": {
+        "define-properties": "1.1.2",
+        "es-abstract": "1.10.0"
+      }
+    },
+    "array-union": {
+      "version": "1.0.2",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/array-union/-/array-union-1.0.2.tgz",
+      "integrity": "sha1-mjRBDk9OPaI96jdb5b5w8kd47Dk=",
+      "dev": true,
+      "requires": {
+        "array-uniq": "1.0.3"
+      }
+    },
+    "array-uniq": {
+      "version": "1.0.3",
+      "resolved": "https://artifactory.secureserver.net/artifactory/api/npm/node-virt/array-uniq/-/array-uniq-1.0.3.tgz",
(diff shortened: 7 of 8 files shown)


LindsayD:
Maybe this is missed timing, but there is pretty extensive discussion of what is being removed and why in the PR description above.

packages/eslint-config-godaddy-minimal/index.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+  extends: [
+    require('./extends')('eslint-config-godaddy-react-flow')
+  ],
+  rules: {
+    'complexity': 0,
```
DullReferenceException: I think having mechanical rules like this are a good thing. Usually "overly complex" is a legitimate concern, and it's easy enough to use a comment to make an exception in the rare case where the complexity is warranted. IMO, this is a case of a linting rule that is actually useful and not just a style preference.

packages/eslint-config-godaddy-minimal/index.js
```diff
@@ -0,0 +1,24 @@
...
+    'consistent-return': 0,
+    'id-length': 0,
+    'indent': [2, 2, {
+      VariableDeclarator: { var: 2, let: 2, const: 3 },
+      flatTernaryExpressions: true,
+      MemberExpression: 'off'
+    }],
+    'max-params': 0,
```
DullReferenceException: Because JavaScript does not have named functions, function calls with too many parameters become unreadable. I have seen far too many code bases where huge parameter lists make reability a nightmare. Also, it's generally bad code smell that a function has too many responsibilities.

"option objects" are far more readable if passing a lot of data is required, since all the pieces of data are named.

packages/eslint-config-godaddy-minimal/index.js
```diff
@@ -0,0 +1,24 @@
...
+    'max-nested-callbacks': 0,
```
DullReferenceException: It's easy enough to turn off the max nested callbacks rule in test folders to prevent the mocha issue where nested callbacks are an OK thing. In regular code, too many nested callbacks is definitely a red flag, so I'd vote to leave this rule in place.

packages/eslint-config-godaddy-minimal/index.js
```diff
@@ -0,0 +1,24 @@
...
+    'max-statements': 0,
```
DullReferenceException: Like the rest of these rules, this one forces you to pause and reconsider your function. Massive "god functions" are pretty much always bad and should be broken down into smaller pieces, so I'd vote to keep this.

chrishinrichs:
Fine with the idea, but I would keep `max-nested-callbacks` and `max-params` personally...I think they're both useful at keeping code manageable and encouraging good style.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: If you read my explanation of the rule in the description, this triggers for any nested function invocations, not just for callbacks. This means that using `map` or similar inside an async callback can push code over the count. We all know better than to nest async control flow too deeply, and this can be easily discussed during code review without limiting higher-order functions in general.

packages/eslint-config-godaddy-minimal/index.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+  extends: [
+    require('./extends')('eslint-config-godaddy-react-flow')
```
jcrugzz: I don't think this gives us a "minimal" starting point

indexzero: Please change this to `eslint-config-godaddy`. From a combinatoric perspective there are a couple of ways to approach this:

1. Disable these rules in `eslint-config-godaddy` and then have another config which is these rules turned on. 
2. Have a `minimal.js` inside each of the other packages that disables these rules, but includes all other rules. 
3. Another approach I haven't thought of.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: I agree that they're bad; that's why they should be discussed in code review. If someone doesn't realize that they're writing a function that's too long, that's a good sign that they should get assistance refactoring it as part of a code review. Hastily slicing a function in half because a tool told someone to do so is likely to lead to an even worse outcome than leaving it intact for reviewers to comment on.

packages/eslint-config-godaddy-minimal/index.js
scommisso: I'm good with keeping this as a warning, as it may help identify areas that did not undergo proper code review

packages/eslint-config-godaddy-minimal/index.js
scommisso: On the fence for this one... I see the argument, but forcing into an "options" object if the function has a large number of arguments is a good thing now that destructuring basically gives us named arguments

packages/eslint-config-godaddy-minimal/index.js
scommisso: I'd like to leave this one in place, with the additional guidance that it should be disabled in unit tests. Not sure if that's something that can be done in this package?

packages/eslint-config-godaddy-minimal/index.js
```diff
@@ -0,0 +1,24 @@
...
+      VariableDeclarator: { var: 2, let: 2, const: 3 },
+      flatTernaryExpressions: true,
+      MemberExpression: 'off'
+    }],
+    'max-params': 0,
+    'max-nested-callbacks': 0,
+    'max-statements': 0,
+    'no-else-return': 0,
+    'no-nested-ternary': 0,
+    'no-shadow': 0,
```
scommisso: I'm on the fence for this one. Accidental shadowing can cause huge headaches, and may be easily missed during code reviews. It is an annoyance during the scenario described above. I'm +1 downgrading this to a warning.

packages/eslint-config-godaddy-minimal/index.js
scommisso: maybe just call this `eslint-config-godaddy-minimal-flow` ?

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: "option objects" are an abstraction built around "I needed all of this data in one place, so I shoved them together". This is not a good abstraction, or something we should encourage. Options objects are also not a general solution, because they complicate type checking and the use of higher-order functions.

If there's a large number of arguments to a function (a function which is exposed to callers, not one which has different sets of default arguments bound to it to produce simpler functions) then the solution is to build better domain abstractions to combine the function's arguments, or more frequently, to divide the function up, separating its responsibilities. Or to do what I just mentioned, and export multiple functions, each of which are the result of binding different default arguments to the main one. There's a lot of wiggle room here, and these are tradeoffs that should be discussed in code review. Making this a lint rule encourages people to bury the problem.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: It's trivial to write overly complex code which doesn't trigger this rule anyway. Measuring cyclomatic complexity can detect a small subset of functions which are too complicated, but does little to nothing to remove responsibility from developers during the code review and refactoring processes. I agree that complex functions are usually bad; but there are cases in which functions can be comprehensible while having high cyclomatic complexity, and these are issues that should be addressed by humans, not mechanically.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: The idea is to be "minimal" in the sense of "permits a wider range of code styles without configuration". Nothing in these rules requires that teams use react or enable flow, but extending this allows teams to use the same style guide whether or not react and flow are in play.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: Is there any real difference between these two, though?
```es6
function doStuff(foo, bar, baz, qux, cat, dog) {
  // ...
}

doStuff(foo, bar, baz, qux, cat, dog);

function doStuff2(foo, bar, baz, { qux, cat, dog }) {
  // ...
}

doStuff2(foo, bar, baz, { qux, cat, dog });
```
Destructuring means that argument count and the apparent number of arguments to a function are no longer strictly coupled; the count of arguments passed is not a good measure of the size or complexity of a function's interface.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: If accidental shadowing is possible, it's a sign that the scope was too large and/or the names being used were too vague, which should both be discussed in code review on their own.

packages/eslint-config-godaddy-minimal/index.js
DullReferenceException: If an inline map function inside of a deep async control flow puts you over the edge, generally the stuff before it was already too deep. 

You keep mentioning that things can be discussed during code review, but it's very easy to miss certain quality metrics during code reviews, and not all code reviewers have similar value systems when it comes to quality. Having some automatic standards is useful for the parts that are generally detectable, and this is one case, IMO.

packages/eslint-config-godaddy-minimal/index.js
DullReferenceException: I disagree that breaking down a complex function "hastily" will normally be a bad thing.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: The fact that this sometimes is helpful and sometimes is harmful is a sign that a minimal set of rules should leave it out, and let teams which want it enable it in their project overrides. If I thought that this rule was harmful 100% of the time I'd be pushing to disable it in the main rule set.

DullReferenceException: There's a big difference when it comes to value passing.

For example, I see stuff like this all the time: `someFunction(app, false, 0, 'test', shopperId);`  What does `false` stand for in this case? Or `0`? I have to look at the source code of that function to understand it, whereas `someFunction({ app, env: 'test', shopperId, recursive: false, maxRetries: 0 })` is much easier to read.

sathify: I agree with jpage. +1 for options object.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: If code reviewers don't have a good, common set of metrics over quality, then the project in question is destined for poor code quality anyway, and the rest of this is moot. It's not possible to build good code without a common understanding of what "good code" is, and this is a much larger and more complex topic than can be handled by syntactic analysis.

Also, re:
> If an inline map function inside of a deep async control flow puts you over the edge, generally the stuff before it was already too deep.
If that's the case, then the existing rule is already failing to catch problem cases, no? And what about a more complex higher-order function:
```es6
someArray.chain(x => someOtherArray.map(y => [x, y]));
```
This is not a heavy-weight function; it just tuples the cartesian product of two arrays. But putting this in a function cuts the total allowed levels of nesting in half. This is a sign that we're mixing concerns: higher-order functions and callbacks for async control flow are two different issues, that should be addressed differently. Linting, by operating at a syntactic level, is not advanced enough to address these issues adequately. Both of these issues can be addressed by competent devs, so this is not a "something is better than nothing" situation.

packages/eslint-config-godaddy-minimal/index.js
fritzmonkey: Seeing it disabled via a comment also serves as a flag during review, making discussion of why it is so complex, and if complexity can be reduced more likely.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: Boolean arguments to a function are a code smell anyway, if you follow the "clean code" stuff. Whether or not this is in an object doesn't change the underlying problem: a function taking a boolean argument is explicitly doing two things, and functions whose interfaces are exposed should only do one thing. Putting this inside an options object hides the fact that this function's responsibility is creeping, and gives an easy way for other people to add even more to it.

packages/eslint-config-godaddy-minimal/index.js
LindsayD: Maybe making this less restrictive than 15 statements would be better. There are some cases where just setting up variable declarations counts as "statements" when there's no way to avoid doing that, and forcing a variation in style to just bunch things with commas instead of declaring each variable separately.  If there was a way to define a statement as something that actually performed an operation, then it would be better, but if not, just making this less restrictive (maybe 20 lines) would help in most cases I've run into.

packages/eslint-config-godaddy-minimal/index.js
LindsayD: Doesn't this also discourage more functional programming, where anonymous callbacks are frequently nested? I think this would be a better PR discussion than a linting rule.

packages/eslint-config-godaddy-minimal/index.js
DullReferenceException: I think it's nonsense to say that booleans always imply that something should be multiple functions. What about two booleans? Is four similar-but-differently-named functions better than one with parameters? Three booleans or 8 named functions? Sometimes is the case, but sometimes booleans are the better option.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: @LindsayD That's why I run into this problem all the time, despite the fact that I never deeply nest async control flow.

fabiomcosta:
This is my list with little comments on each lint.

**Enable:**
* consistent-return
* no-shadow
* no-unused-vars
* spaced-comment
* no-else-return
* max-params
* max-nested-callbacks

**Disable:**
id-length
no-nested-ternary
no-undefined

**Don't care:**
indent

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: I am good with that as a compromise.

packages/eslint-config-godaddy-minimal/index.js
JosephJNK: And now we've gotten to the point where we're discussing what the best interface for something should be at the level of engineers, which is where all discussions about interfaces should take place. None of the questions you raise here can be addressed via syntactic analysis.

DullReferenceException: Considering my example of unreadable function call versus more readable, I think having the linting rule remains a better idea than disabling it.

LindsayD: This may be a case of loosening the restriction a bit. 4 params is extreme, 6 would be better and still get the point across. 

I don't agree that having an object parameter with a ton of properties is any more readable or less of a code smell than a parameter list of the same number. Not saying a lot of params/properties is a good thing, but not understanding how wrapping them in an object makes it any better, plus wrapping forces you to have to change the caller's signatures wherever they may be if a new parameter is an enhancement to the function.

fabiomcosta:
This is my list with little comments on each rule.

**Enable:**
* no-shadow - it can detect real bugs on development time and it's generally easy to work around it in cases you know what you're doing.
* no-unused-vars - Yes, let's keep it enabled with `ignoreRestSiblings` set to true.
* spaced-comment - I like that this can be statically checked and that new code is forced to follow the same pattern. 
* no-else-return - I like that this can be statically checked and that new code is forced to follow the same pattern.
* max-params - As already pointed out this is generally a sign of bad API/function signature.
* max-nested-callbacks - As already pointed out this is generally a sign of bad code that needs refactoring.
* consistent-return - Is unnecessary on projects using a static analysis tool. But what about the other projects?
* no-nested-ternary - Depends on how you write them. I've been annoyed by this in the past but I think this one is better on than off.

**Disable:**
* complexity - As pointed out by the PR description, defining complexity of a function is a gray line. People are generally ok with this check until it hits them on a function that looks perfectly normal.
I can clearly see both sides on this one... but I consider the hard complexity value too arbitrary.
* max-statements - kind of the same thing as complexity.
* id-length - The id should have any size it needs to be understandable to the reader
* no-undefined - There is no reason to have this on.

**Don't care:**
* indent

packages/eslint-config-godaddy-minimal/index.js
```diff
@@ -0,0 +1,24 @@
...
+    'indent': [2, 2, {
+      VariableDeclarator: { var: 2, let: 2, const: 3 },
+      flatTernaryExpressions: true,
+      MemberExpression: 'off'
+    }],
+    'max-params': 0,
+    'max-nested-callbacks': 0,
+    'max-statements': 0,
+    'no-else-return': 0,
+    'no-nested-ternary': 0,
```
LindsayD: Want to add a +1 for disabling this rule, one of my biggest pet peeves.  Ternaries can be used to make code a lot more readable compared to cluttering if/then/else chains or switch statements.  Having to constantly disable this rule is a pain.

packages/eslint-config-godaddy-minimal/index.js
```diff
@@ -0,0 +1,24 @@
...
+    'no-shadow': 0,
+    'no-undefined': 0,
+    'no-unused-vars': ['error', { args: 'none', ignoreRestSiblings: true }],
+    'spaced-comment': 0
```
LindsayD: If we don't remove this rule, can it just be one that gets automatically fixed like some other formatting is, or a warning? I have never understood why such a trivial thing is a linting `error`.  And it is so annoying when you just want to comment out several lines while actively testing code to have the linter error because of a space.

packages/eslint-config-godaddy-minimal/index.js
LindsayD: Though it's probably rare, there's a chance wrapping params in an object where the object could change during execution could cause bugs:
```
const p = { one: 1 };

function sideEffect(params) {
    console.log(params.one);
    p.one = 2;
    console.log(params.one);
}

sideEffect(p); // renders 1 and then 2
```
Using destructuring, immutability and a few other things would also prevent that, but it's still another potential consequence of forcing param wrapping.

scommisso: `function doStuff2(foo, bar, baz, { qux, cat, dog }) {`
the horror!

Honestly, function signatures are the things easily catchable in code reviews. A huge function signature, or a monstrosity like that above would get -1'd pretty quick.

As this style guide is defined as a "minimal" guide, I'm +1 on removing that rule.

packages/eslint-config-godaddy-minimal/index.js
scommisso: Eh, yeah I see that. Shadowing is harder to catch in code reviews since the original declaration might not be visible in the GitHub reviewer (they'd have to click to expand upwards and/or downwards), in the case where existing code is being modified.

But, in keeping that this is a new "minimal" style guide, I'm OK with this. There will be potential problems either way.

ahernandez-miralles:
I +1 other comments about turning some of these rules into warnings instead of outright disabling them. Not all dev teams are able to have all PRs code reviewed by a senior engineer (due to lack of resources), so IMO having the linter generate warnings would be a great way to spark discussion around code quality for more junior teams, and it still wouldn't block development.

JosephJNK:
@ahernandez-miralles It sounds like your team would be better suited by the main style guide than by a minimal style guide, then. This PR isn't changing the main set of rules, but rather creating a less opinionated ruleset consisting only of rules which are relatively uncontested across teams. (uncontested as measured by degree of developer objections and frequency with which the rules are disabled)

indexzero:
Closing due to age. Feel free to re-open if strong feelings remain. Thanks all for the lively discussion.

PR #59, opened 2017-12-14
Update ecmaVersion to 8

bjisa:
Hey there, any reason we can't update ecmaVersion to 8? I'm using async/await, and currently have to override ecmaVersion for eslint to parse these keywords.


rkmarks:
👍

PR #60, opened 2017-12-21
feat(lint): prep for semantic release by enforcing commit style

gergelyke:
This PR preps the repo for adding semantic release, by enforcing the angular commit style.

*But who lints the linters? 🤔*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 sudo: false
 language: node_js
 node_js:
-  - "6.9"
+  - "8.9.3"
 
 script:
+  - npm run commitmsg:ci
   - npm test
 
 matrix:
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,1 @@
+module.exports = {extends: ['@commitlint/config-angular']};
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1091 @@
+{
+  "name": "godaddy-style",
+  "version": "5.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@commitlint/cli": {
+      "version": "5.2.5",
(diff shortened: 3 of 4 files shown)


rkmarks:
Are we commitng the package-lock.json files now?

3rd-Eden:
@rkmarks yes, we should be commiting those to git. Ideally with `.gitattributes` setting it as `binary` file so it doesn't cause merge conflicts.

3rd-Eden:
I honestly don't see the need for linting commit messages here. It only creates a bigger barrier for contributors and it doesn't really add anything to the project. I don't see the reason why we should allow `chore(lint) added linter` vs `[pkg] Added linter` Both achieve the same goal, except for the last one you get yelled at and will be prevented to commit.

3rd-Eden:
Also if the commit messages of a contributor is up to spec, you can just do a [Squash and merge] on github and alter the commit message.

indexzero:
@3rd-Eden I am relatively new to semantic release, but if I understand it correctly commit messages have to be automatically sanitized to a format so semantic release can understand when to do a patch vs. minor vs. major bump.

I too am not sold on the Angular commit style. It's the holidays – if it's all the same to everyone else would be great to think on this and revisit it in the New Year.

3rd-Eden:
Regardless of the style used, I feel that declining commits because it doesn't satisfy a certain pattern is a bad pattern. I'm fine with steering people to use a given format. Instead of linting, adoption a [commit.template](https://robots.thoughtbot.com/better-commit-messages-with-a-gitmessage-template) would be more helpful. 

For example, the [pre-commit](https://github.com/observing/pre-commit) allows you specify a template that will be the initial body of the commit which users can fill in.

gergelyke:
Most probably, we don't need the linter at all, and just opt for the analyzeCommit plugin for semantic-release: https://github.com/semantic-release/semantic-release#analyzecommits

PR #62, opened 2018-01-31
Enable jsx-equals-spacing

fabiomcosta:
Enables `react/jsx-equals-spacing`. The dafault value is "never", so basicaly:

`<Hello name = {firstname} />;` becomes `<Hello name={firstname} />;`

See full doc: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-equals-spacing.md


fabiomcosta:
@indexzero @gergelyke

fabiomcosta:
ping, I can't merge it, I'm not an admin

PR #67, opened 2018-07-25
fix(cli): fix no-shadow warning

yungcheng:
`err` is already declared in line#28  `fs.readdir(cwd, function (err, files) {`

### DIFF
--- a/packages/eslint-config-godaddy/cli.js
+++ b/packages/eslint-config-godaddy/cli.js
@@ -36,8 +36,8 @@ module.exports = function (filename) {
       process.argv.splice(2, 0, '-c', require.resolve(filename));
     }
 
-    which('eslint', function (err, resolved) {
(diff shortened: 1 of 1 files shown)


indexzero:
Personally I am -1 on this, but interested in what other folks think. `no-shadow` is a particularly divisive rule amongst folks, which is why it is a warning. Warnings by definition need not be fixed (otherwise they would be errors).

rkmarks:
👎 I too am against this as they should be a warning to the developer/team to be fixed. This doesn't mean it will necessarily break the code however it can cause additional confusion for other developers  who are working on the code base. Making it more strict, aka an error, signifies that something is breaking IMO. I can see both sides of the aisle on this but a warning is more justified than an error.

indexzero:
Closing due to age.

PR #68, opened 2018-07-25
feat(rule): object-curly-spacing never

yungcheng:
Proposed rule change: `object-curly-spacing` from `always` to `never`, also 
`react/jsx-curly-spacing` from `always` to `never`

The rules we have now it's not consistent, we set `object-curly-spacing` to `always` but with `objectsInObjects` set to `false`, which makes the following valid
```js
/*eslint object-curly-spacing: ["error", "always", { "objectsInObjects": false }]*/

var obj = { "foo": { "baz": 1, "bar": 2 }};
```
(Note the spacing is asymmetric considering in the end there's no space, but there is a space after opening curly brace)

Same for our `react/jsx-curly-spacing` rule, we set it to `always` but `spacing: { objectLiterals: 'never' }`, which makes the following valid:
```jsx
<App blah={ 3 } foo={{ bar: true, baz: true }} />;
```

This PR turns both rule to `never` which eliminates all unnecessary spaces and confusions. It also makes code more compact and easier to read and author.

References:
[object-curly-spacing](https://eslint.org/docs/rules/object-curly-spacing#never)
[react/jsx-curly-spacing](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-spacing.md#never)

### DIFF
--- a/packages/eslint-config-godaddy-react/index.js
+++ b/packages/eslint-config-godaddy-react/index.js
@@ -12,7 +12,7 @@ module.exports = {
   },
   rules: {
     'react/display-name': 0,
-    'react/jsx-pascal-case': [2, { allowAllCaps: true }],
+    'react/jsx-pascal-case': [2, {allowAllCaps: true}],
     'react/jsx-uses-react': 1,
     'react/jsx-equals-spacing': 2,
     'react/prefer-es6-class': 2,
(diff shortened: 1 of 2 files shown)


indexzero:
Thoughts on simply changing `objectsInObjects` to `true`?

yungcheng:
close this now that #48 is merged as the spacing is now symmetric

PR #70, opened 2018-10-26
Get rid of old lodash

DullReferenceException:
We received a security vulnerability alert for older versions of lodash. Upgrade packages until that bad version is eliminated.


indexzero:
Post-humous 👍

PR #73, opened 2018-12-11
[major] depend on latest babel-eslint for react

Swaagie:
Latest `eslint` will require `parser: 'babel-eslint'` for code containing JSX. Update `eslint-config-godaddy-react` to use this by default and update deps. Since `react-flow` was already using these dependencies update them alongside.


packages/eslint-config-godaddy-react/package.json
```diff
@@ -28,17 +28,19 @@
     "eslint-config-godaddy": "^2.0.0"
   },
   "peerDependencies": {
-    "eslint": "^4.3.0",
+    "babel-eslint": ">=7.2.1 <11.0.0",
```
msluther: why is this? to stop it from needing babel@7?

indexzero: It's just being flexible since it is backwards compatible with older versions of `babel-eslint`. There is no need to be prescriptive to a `^` versions. See `packages/eslint-config-godaddy-react-flow/package.json` above which also uses this.

fabiomcosta: Hmm, why not just `>=7.0.0`?

PR #76, opened 2019-02-11
Add eslint-plugin-jsx-a11y

SivanMehta:
Adds [`eslint-plugin-jsx-a11y`](https://www.npmjs.com/package/eslint-plugin-jsx-a11y) to default `React` linting.

### DIFF
--- a/packages/eslint-config-godaddy-react/index.js
+++ b/packages/eslint-config-godaddy-react/index.js
@@ -1,11 +1,10 @@
 module.exports = {
   extends: [
     require('./extends')('eslint-config-godaddy'),
-    'plugin:eslint-plugin-react/recommended',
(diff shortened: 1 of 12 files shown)


packages/eslint-config-godaddy-react/index.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   extends: [
     require('./extends')('eslint-config-godaddy'),
     'plugin:eslint-plugin-react/recommended'
+    'eslint-plugin-jsx-a11y'
```
kinetifex: Shouldn't this be `'plugin:jsx-a11y/recommended'` here for extend?

SivanMehta: I guess? Not sure how to test given that this is the test command for this subpackage:

```
"test": "echo ok",
```

PR #84, opened 2019-08-13
Eslint 6 support (major)

DullReferenceException:
- Enable `eslint@6`
- Fix peerDependencies to align with devDependencies
- Remove unused dependency

### DIFF
--- a/packages/eslint-config-godaddy/package.json
+++ b/packages/eslint-config-godaddy/package.json
@@ -24,16 +24,15 @@
     "eslint-godaddy": "bin/eslint-godaddy"
   },
   "peerDependencies": {
-    "eslint": "^5.8.0",
-    "eslint-plugin-json": "^1.2.0",
-    "eslint-plugin-mocha": "^5.2.0"
+    "eslint": "^6.1.0",
+    "eslint-plugin-json": "^1.4.0",
+    "eslint-plugin-mocha": "^5.3.0"
   },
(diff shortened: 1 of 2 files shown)


packages/eslint-config-godaddy/package.json
```diff
@@ -24,16 +24,15 @@
     "eslint-godaddy": "bin/eslint-godaddy"
   },
   "peerDependencies": {
-    "eslint": "^5.8.0",
-    "eslint-plugin-json": "^1.2.0",
-    "eslint-plugin-mocha": "^5.2.0"
+    "eslint": "^6.1.0",
```
msluther: Should this change across all the packages at once?

DullReferenceException: I'm not sure if I can do that, sadly. The problem is that some of these packages have dependencies on other packages in this repo, and those packages have `eslint` peer dependencies which would be disallowed. So I'm going to have to publish this one first, then update those dependent on it afterward (in fact, I've got one queued up that does just that).

scommisso: Can do a major version bump of this library, then consumers who want ESlint5 can remain on the previous version, right?

DullReferenceException: @scommisso: that's the plan, to make this a major bump (going to do that post-merge).

PR #112, opened 2020-07-07
Force == for null to catch unhandled undefined

subtleGradient:
Why use `== null` vs `=== null || === undefined` (or `x === null || typeof x === 'undefined'`)  
when they technically behave exactly the same?

## One argument is **brevity**

`== null` uses fewer bytes across the wire.
Less code is always better code. Fewer possible bugs.
Code is more readable, etc...
Sure

But a much more important reason is...

## catch subtle edge case bugs

When treating `null` and `undefined` separately, it doubles the cognitive load of every `if` check in the codebase. It's extra work to go through all code and double check that we're checking for both cases. That can't be enforced with a lint rule so it needs to be part of the manual review process.

e.g. two bugs were easy to fix by enforcing `== null` always.

But what would we need to do to go the other way? 
There's no automatic way to check for a missing `=== undefined` check. 
So we would need to manually review everywhere `null` is used throughout the codebase.


rxmarbles (approved):

rxmarbles:
@subtleGradient  would you mind updating from latest when you get a chance. I'll get another eye on this PR for visibility

msluther (approved):

PR #127, opened 2020-10-03
make readme example match readme instructions

thumbsupep:
ESLint supports configuration files in several formats. The readmes instructed to define `.eslintrc` but gave the example code for `.eslintrc.js`
This PR makes the instruction and example consistent.
https://eslint.org/docs/user-guide/configuring#configuration-file-formats

### DIFF
--- a/packages/eslint-config-godaddy-es5/README.md
+++ b/packages/eslint-config-godaddy-es5/README.md
@@ -35,7 +35,7 @@ These use _exactly_ the configuration defined in this  package (`eslint-config-g
 ##### 2. Define your local `.eslintrc` and run `eslint` yourself:
 
 ``` js
-module.exports = {
+{
   // Or for ES5 + React:
   // extends: 'godaddy-es5-react',
   extends: 'godaddy-es5',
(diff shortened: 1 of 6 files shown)


packages/eslint-config-godaddy-es5/README.md
```diff
@@ -35,7 +35,7 @@ These use _exactly_ the configuration defined in this  package (`eslint-config-g
 ##### 2. Define your local `.eslintrc` and run `eslint` yourself:
 
 ``` js
-module.exports = {
+{
   // Or for ES5 + React:
   // extends: 'godaddy-es5-react',
   extends: 'godaddy-es5',
```
msluther, line 41: This isn't technically correct either since:

1. `.eslintrc` is required to be a json file and
2. JSON files can't contain comments
3. JSON requires the property names to be quoted
4. JSON requires the string values to use double quotes `"`
5. The markdown code-block language selector should be `json` instead of `js`

thumbsupep: fixed

msluther (approved):

PR #128, opened 2020-10-28
Update travis ci

rxmarbles:
This PR removes Node@8 as it is no longer maintained and adds Node@12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "8"
   - "10"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,8 +1,8 @@
 sudo: false
 language: node_js
 node_js:
-  - "8"
   - "10"
+  - "12"
```
msluther, line 5: Perhaps 10/12/14, since 14 is now LTS.

msluther (approved):

PR #129, opened 2020-12-17
Migrate to GH Actions

rxmarbles:


tarkatronic (approved):

package.json
```diff
@@ -19,14 +19,11 @@
...
-  },
-  "prepush": {
-    "tasks": [
-      "npm run prepush"
-    ]
+    "test:config:react-typescript": "cd packages/eslint-config-godaddy-react-typescript && npm test"
   },
+  "pre-commit": [
+    "test"
+  ],
```
msluther, line 26: I'm not a fan of this, it discourages certain styles of development (e.g. TDD). why shouldn't I be able to commit locally with broken tests. Yeah I can always do `--no-verify` and commit anyway, but its a thing I need to remember. 

What exactly are we trying to prevent with this? direct commits to master with broken tests? PRs being created with broken tests (I actually think its useful sometimes to have this)?

silasbw: I don't really have a horse in this race, but I'm +1 to @msluther's feedback.

rxmarbles: that's fair, though tbh it was `prepush` before. I just moved away from `prepush` due to a security vuln. If we want to remove it I'm all for it and can probably be done in a follow up PR where we can get a full consensus on it.

msluther: If we want to break it up, can we have another PR just drop it and try to merge that before this, rather than touching it to change what its doing int the GHActions PR?  I'm also okay just dropping it here if we get enough folks to weigh in on it.

tarkatronic: I agree with the approach of reverting this to its original behavior and having a follow-up PR to address it. This really isn't related to CI, so should be handled and discussed separately.

msluther: https://github.com/godaddy/javascript/pull/130

samshull: I'm trying to wrap my head around this. The argument is that it is better to rely on github to ensure that tests are run rather than local git because `--no-verify` is too onerous? Is that correct or am I misunderstanding?

Also, couldn't we just replace [pre-push](https://www.npmjs.com/package/pre-push) with [prepush](https://www.npmjs.com/package/prepush) instead of switching to pre-commit?

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "6.0.0",
   "description": "The approach to JavaScript at GoDaddy. We think it's pretty decent.",
   "scripts": {
+    "prepublishOnly": "npm test",
```
DullReferenceException, line 6: I'm not sure this will do much since we don't publish at the top level but instead on an individual package basis.

tarkatronic: Good callout. Is it possible for this to be placed in the sub-packages instead?

msluther: Can we do this in a separate PR?

rxmarbles: Fair enough I'll do a separate PR to script it out better for publishing and such

msluther (approved):

PR #130, opened 2020-12-17
Remove prepush hooks

msluther:
From this discussion: https://github.com/godaddy/javascript/pull/129#discussion_r545266041

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,13 +19,7 @@
     "test:config:react": "cd packages/eslint-config-godaddy-react && npm test",
     "test:config:react-flow": "cd packages/eslint-config-godaddy-react-flow && npm test",
     "test:config:typescript": "cd packages/eslint-config-godaddy-typescript && npm test",
-    "test:config:react-typescript": "cd packages/eslint-config-godaddy-react-typescript && npm test",
-    "prepush": "npm test"
-  },
-  "prepush": {
-    "tasks": [
-      "npm run prepush"
-    ]
+    "test:config:react-typescript": "cd packages/eslint-config-godaddy-react-typescript && npm test"
   },
   "repository": {
     "type": "git",
@@ -68,7 +62,6 @@
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-mocha": "^5.3.0",
     "eslint-plugin-react": "^7.14.2",
-    "prepush": "^3.1.11",
     "react": "^16.8.6",
     "typescript": "^3.0"
   },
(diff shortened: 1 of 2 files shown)


rxmarbles (approved):

tarkatronic (approved):

samshull (changes requested):
I would like to better understand why removal of this without a replacement is the best choice. Could you clarify for me, please?

package.json
```diff
@@ -68,7 +62,6 @@
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-mocha": "^5.3.0",
     "eslint-plugin-react": "^7.14.2",
-    "prepush": "^3.1.11",
```
msluther, line 71: @samshull Random comment location here to allow threaded messages in github.

1. `prepush` has security vulnerability and was being replaced in another PR (linked in this PR description) with a precommit hook.
2. Why dictate a users's workflow or force them to use `--no-verify`.  The proper time to verify what someone has done is passing tests is at PR/merge time and no on every commit or push. If a developer wants to commit frequently small chunks of code, that should actually be encouraged. If a developer wants to use TDD and commit an initial set of breaking tests, that should be encouraged/allowed. If a developer wants to push frequently to avoid losing their work, they should be allowed to do so.

tarkatronic: I think doing the way proposed here (no prepush) is more of a "trust but verify" model, which is generally a good thing. You encourage the developer to run the test suite, and trust that they do when necessary. And then when a PR is created, you verify that it passes by running it again.

samshull: There does not appear to be master branch protections on this repo. If we are to rely on PR testing, we should start by changing those rules first. And while I would prefer not to be a devil's advocate here, neither push, nor PR testing will prevent this package from being published to npm with bad code, so I would argue that we should be more concerned with prepublish actions than either of these.

tarkatronic: prepublish sounds like a great idea! Would you be willing to spin up a PR to propose that? :)

tarkatronic: As for branch protection rules, I just got those started with 2 required reviews; once the GH Actions PR gets merged, we can also set those up as required. 👍🏻

DullReferenceException: (remember to use `prepublishOnly` to avoid the silly npm behavior of running `prepublish` after _installs_)

mswaagman-godaddy (approved):

samshull (commented):
I don't care whether we enforce testing with push hooks or github actions, as long as it accomplishes the goal of ensuring that code in the master branch is passing the tests we have written.

PR #131, opened 2021-02-26
**Major** Add react-hooks lint rules to react rules

msluther:
The react-hooks rules catch lots of errors when using `useMemo`/`useEffect`/`useCallback` especially around the dependency arrays.  e.g. 

```
  35:6  warning  React Hook useEffect has a missing dependency: 'props'. Either include it or remove the dependency array. However, 'props' will change when *any* prop changes, so the preferred fix is to destructure the 'props' object outside of the useEffect call and refer to those specific props inside useEffect  react-hooks/exhaustive-deps
  35:6  warning  React Hook useEffect has a missing dependency: 'foo'. Either include it or remove the dependency array.   react-hooks/exhaustive-deps
```

But, because we would be including new rules, this should go out as a new Major release for the react* rules.


DullReferenceException (approved):
💯

rxmarbles (approved):

twyrick-godaddy (approved):
👍

fritzmonkey (approved):

wjarvis-godaddy (approved):

runruh-godaddy (approved):

DullReferenceException:
:shipit:

rxmarbles:
Will merge it now and ship later tonight

PR #133, opened 2021-03-17
Convert to lerna mono-repo

rxmarbles:
Due to issues w/ the current setup in #131 I recommend moving to a better mono-repo setup for now (until npm workspaces is in a better situation)


packages/eslint-config-godaddy-es5/package.json
```diff
@@ -27,7 +27,8 @@
     "eslint-godaddy-es5-react": "bin/eslint-godaddy-es5-react"
   },
   "dependencies": {
-    "eslint-config-godaddy": "^3.0.0"
+    "eslint-config-godaddy": "^3.0.0",
+    "eslint-config-godaddy-react": "^5.0.0"
```
DullReferenceException, line 31: I suggest we eliminate this es5 package; if someone really needs to code in ES5 they can use an old release.

DullReferenceException (approved):

packages/eslint-config-godaddy-es5/package.json
rxmarbles: I'm also in favor of this tagging @godaddy/javascript for additional feedback on this

packages/eslint-config-godaddy-flow/package-lock.json
```diff
@@ -494,16 +494,6 @@
         }
       }
     },
-    "eslint-config-godaddy": {
```
msluther, line 497: Should we also kill the flow package? Are folks still using it?

DullReferenceException: Sadly yes, in my org :(

lerna.json
```diff
@@ -0,0 +1,6 @@
+{
+  "packages": [
+    "packages/*"
+  ],
+  "version": "independent"
```
msluther, line 5: /random location to allow threaded messages.

We should probably get some README.md updates for how to dev locally with lerna / how to publish / etc. And if there are any recommendations for majors (let them major bump independently vs major bumping everything together).

asilvas (approved):

lerna.json
rxmarbles: Good call out, the good part about independent is that we can for sure no longer bump something that didn't get a change. I will make a some README changes tonight. I got tied up w/ other things on my plate 😛

rxmarbles: Updated readme w/ dets. Going to merge as I also fixed a critical security patch form dependabot

PR #156, opened 2021-12-15
Eslint v8

kawikabader:
Updates made to support using Eslint v8.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,11 @@
     "bootstrap": "lerna bootstrap --hoist --ignore-scripts",
     "postinstall": "npm run bootstrap",
     "test": "lerna run test --stream",
-    "publish": "lerna publish"
+    "publish": "lerna publish",
+    "clean": "npm run clean:artifacts && npm run clean:packages && npm run clean:root",
+    "clean:artifacts": "lerna run clean --parallel",
+    "clean:packages": "lerna clean --yes",
+    "clean:root": "rimraf node_modules"
   },
   "repository": {
     "type": "git",
@@ -40,13 +44,14 @@
     "godaddy-js-style-eslint": "./bin/eslint"
   },
   "devDependencies": {
-    "eslint": "^7.25.0",
+    "eslint": "^8.4.1",
     "eslint-plugin-json": "^2.1.2",
     "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-mocha": "^6.3.0",
     "eslint-plugin-react": "^7.23.2",
     "lerna": "^4.0.0",
     "react": "^17.0.2",
+    "rimraf": "^3.0.2",
     "typescript": "^4.2.4"
   },
   "dependencies": {}
--- a/packages/eslint-config-godaddy-flow/package.json
+++ b/packages/eslint-config-godaddy-flow/package.json
@@ -36,6 +36,7 @@
     "eslint-plugin-flowtype": "^8.0.3"
   },
   "devDependencies": {
-    "eslint": "^8.4.1"
+    "eslint": "^8.4.1",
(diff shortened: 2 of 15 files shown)


packages/eslint-config-godaddy-flow/package.json
```diff
@@ -29,9 +29,13 @@
     "eslint": ">=6.0.0"
   },
   "dependencies": {
-    "babel-eslint": "^10.0.3",
+    "@babel/core": "^7.16.5",
```
jpage-godaddy, line 32: Could this be a peer + dev dependency instead? Duplicated `@babel/core` packages can be a problem. If moving to `peerDependencies`, I'd also suggest making it less restrictive, like `^7.0.0` in case the consumer has to be stuck on an older `@babel/core` for tragic reasons.

packages/eslint-config-godaddy-react-flow/package.json
```diff
@@ -30,10 +30,11 @@
     "eslint": ">=6.0.0"
   },
   "dependencies": {
-    "babel-eslint": "^10.0.3",
+    "@babel/core": "^7.16.5",
```
jpage-godaddy, line 33: Same feedback as elsewhere; we should probably not have `@babel/core` as a direct dependency because then eslint and the app may have conflicting copies installed.

package.json
```diff
@@ -40,13 +44,14 @@
     "godaddy-js-style-eslint": "./bin/eslint"
   },
   "devDependencies": {
-    "eslint": "^7.25.0",
+    "eslint": "^8.4.1",
```
jpage-godaddy, line 47: Will we want a global `peerDependencies` bump for `eslint` to be on `^8` in case the plugin bumps are not compatible with older versions of `eslint`?

packages/eslint-config-godaddy-flow/package.json
```diff
@@ -29,9 +29,13 @@
...
+    "@babel/eslint-parser": "^7.16.5",
```
msluther, line 33: is the new parser backwards compatible with eslint@6 & eslint@7?

We have this:

```json
  "peerDependencies": {
    "eslint": ">=6.0.0"
  },
```

msluther (commented):
This feels like a new major rev to me, are there other things we'd want to do before publishing? Do we want a new minimum version of eslint? Have we tested this with using older versions?

packages/eslint-config-godaddy-flow/package.json
kinetifex: If the new parser is _not_ compatible with 6 & 7, we should probably have a patch release for our current version to restrict eslint to less than 8 `"eslint": ">=6.0.0 <8.0.0"`. Then include the new parser changes in a new major release.

kawikabader: @msluther `babel-eslint` was [deprecated in March 2020](https://tjaddison.com/blog/2021/03/updating-babel-eslint-to-babeleslint-parser-for-react-apps/#:~:text=As%20of%20March%202020%2C%20babel,%40babel%2Feslint%2Dparser%20.) and eslint@7 was [released on May 2020](https://github.com/eslint/eslint/releases/tag/v7.0.0), so we shouldn't have an issue with compatibility for eslint@6 and 7.

package.json
kawikabader: Just so I understand, you are suggesting that we add eslint@8 as a peer dep at the root package.json?

jpage-godaddy: No, to individual packages (unless adding to root does some magic and adds to each package)

jpage-godaddy (approved):
Besides the `peerDependencies` suggestion for eslint, looks good to me.

kinetifex (approved):

PR #238, opened 2022-12-12
feat: upgrade config-godaddy parser option

mmason2-godaddy:


packages/eslint-config-godaddy/index.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   plugins: ['mocha', 'json'],
   extends: ['eslint:recommended'],
   parserOptions: {
-    ecmaVersion: 9,
+    ecmaVersion: 13,
```
agerard-godaddy, line 13: Wow, this was pretty far behind.

agerard-godaddy (approved):

mmason2-godaddy:
I can't merge just FYI

packages/eslint-config-godaddy/index.js
agerard-godaddy: Actually, before we merge... is ecmaVersion 13 available on eslint v7? We may need to bump the peerDependencies and as a result require a new major release of this config

rmarkins-godaddy: @agerard-godaddy @mmason2-godaddy just as a follow up I am going to bump the `peerDependency` to v8 prior to release of a new major version

PR #974, opened 2024-12-21
chore(eslint)!: migrate all eslint configs to v9

rxmarbles:
## Summary

This PR upgrades all eslint configs to use v9. I know that in eslint >= 8.23 flat config is supported however, feasibly I do not want to make these config semi backwards. Better to do the breaking change now and just keep a v8 branch for patch/bug updates.  I've also removed the `flow` based configs. I know Flow is still around and utilized but we no longer support its use internally at GoDaddy. If there are hard opinions to this I do not mind bringing them back in this PR and migrating them to v9. I've also remove the CLI portion of these libraries as we should only be using the `eslint` binary directly. 

The biggest question here is should be using all the custom rules we have established here or should we be leveraging the industry standards. A lot of our rules are outdated and often times conflicted with a lot of `recommended` configs and plugins. IMHO we should probably lean that way and allow the consumer to decide if they want to modify the rules they want to have. The config should mostly provide the additional configs and plugins (such as the react, mocha, node, ect.) doing this keep their configs small but still allows for flexibility. 

## Changeset

## Test Plan

These have been tested locally to make


.changeset/blue-rats-notice.md
```diff
@@ -0,0 +1,10 @@
+---
+"eslint-config-godaddy-react-typescript": major
+"eslint-config-godaddy-typescript": major
+"eslint-config-godaddy-react": major
+"eslint-config-godaddy": major
+---
+
+chore(eslint)!: migrate all eslint configs to eslint@v9
+
+BREAKING CHANGE: removes v8 support and transistions configs to flat config
```
3rd-Eden, line 10: Might want to explicitly call out that these are now `type:module` instead of the previous `cjs` format.

3rd-Eden (approved):

3rd-Eden:
>  know Flow is still around and utilized but we no longer support its use internally at GoDaddy.

There's 3 projects still depending on this package internally, but I agree, that it's not enough to warrent keeping this configuration around.

cburton-godaddy:
> I know Flow is still around and utilized but we no longer support its use internally at GoDaddy

The whole of ReComm (HostEurope, ex 123-Reg, Heart Internet) uses Flow, must be over 20 Node.js/React applications. These applications are maintenance-only though, so this change does not concern me.

packages/eslint-config-godaddy/package.json
```diff
@@ -3,37 +3,39 @@
...
-    "eslint",
-    "es6"
+    "eslint"
   ],
-  "bin": {
-    "eslint-godaddy": "bin/eslint-godaddy"
-  },
   "peerDependencies": {
-    "eslint": "^8"
+    "eslint": ">= 9"
```
cburton-godaddy, line 27: **question (non-blocking):** I presume `npm` does not care about about the whitespace between `>=` and `9`, but just raising it here.

cburton-godaddy (approved):

kbader-godaddy (approved):

packages/eslint-config-godaddy-react-typescript/index.js
```diff
@@ -1,38 +1,43 @@
...
     },
-    {
-      files: ['*.ts', '*.tsx'],
-      rules: {
-        'react/prop-types': 'off'
-      }
+    plugins: {
+      '@typescript-eslint': tseslintPlugin
+    },
+    files: ['*.ts', '*.tsx'],
```
jpage-godaddy, line 13: There are also variations for TypeScript files that produce `.cjs` and `.mjs` files.

```suggestion
    files: ['*.ts', '*.tsx', '*.mts', '*.mtsx', '*.cts', '*.ctsx'],
```

packages/eslint-config-godaddy-react-typescript/index.js
```diff
@@ -1,38 +1,43 @@
...
+          ignoreRestSiblings: false
+        }
+      ],
+      // Disable the base rule and enable TypeScript version to avoid incorrect reports
+      // See https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-shadow.md
+      'no-shadow': 'off',
+      '@typescript-eslint/no-shadow': 'warn'
+    }
+  },
+  {
```
jpage-godaddy, line 35: Why do we have an array of configs here? It's not clear why we have two different configs for TypeScript.

packages/eslint-config-godaddy-typescript/index.js
```diff
@@ -1,31 +1,37 @@
...
+const config = [
+  ...gdConfig,
+  {
+    languageOptions: {
+      parser: tseslintParser
+    },
+    plugins: {
+      '@typescript-eslint': tseslintPlugin
+    },
+    files: ['*.ts'],
```
jpage-godaddy, line 14:
```suggestion
    files: ['*.ts', '*.mts', '*.cts'],
```

packages/eslint-config-godaddy/package.json
```diff
@@ -3,37 +3,39 @@
...
     "pretest": "npm run --silent lint",
     "test": "echo ok",
-    "clean": "rimraf node_modules",
-    "unused": "eslint-find-rules --unused ./index.js || echo ''"
+    "clean": "rimraf node_modules"
   },
   "repository": {
     "type": "git",
-    "url": "git@github.com:godaddy/javascript.git"
+    "url": "git@github.com:godaddy/javascript.git",
```
jpage-godaddy, line 16: Side note... we really ought to rename this repo.

rxmarbles: Agreed though naming is hard sometimes 😅

packages/eslint-config-godaddy/package.json
agerard-godaddy: I recommend sticking to the major version range. Changes are there that will break and change again in a future 10.x.

```suggestion
    "eslint": "^9"
```

packages/eslint-config-godaddy/index.js
```diff
@@ -1,158 +1,170 @@
...
+          globalReturn: true
+        }
+      }
+    },
+    plugins: {
+      jsdoc,
+      json,
+      mocha
+    },
+    rules: {
```
agerard-godaddy, line 28: > The biggest question here is should be using all the custom rules we have established here or should we be leveraging the industry standards. A lot of our rules are outdated and often times conflicted with a lot of recommended configs and plugins. IMHO we should probably lean that way and allow the consumer to decide if they want to modify the rules they want to have. 

We can churn on rules and debate which ones are "industry standard" for weeks. I suggest wrapping up and merging this PR as a first step towards v9 compatibility. This PR does a great job flattening for v9 and cleaning up the old CLI stuff 👍 

If you have found some preconfigured `recommended` ruleset from that correspond close to what we have today, then PR those separately to focus the discussion. I would avoid any drastic rule changes however, like switching to "standard" or something like that. But if we can consolidate to using more `recommended` configures from our existing dependencies (`eslint-plugin-react`, `eslint-plugin-json`, etc) I could get behind that.

rxmarbles: @agerard-godaddy I agree. I'll get your last comments in and merge this PR so we can look to progress on `rules`

agerard-godaddy (approved):

PR #1036, opened 2025-04-17
fix(globals): Add globals to shared root config

rmarkins-godaddy:
## Summary

This PR adds the `globals` package that was setup by default in v8 and was removed in v9. I've also updated the README's to reference flatpack config and updated the deps for these to latest minors. I did update turbo to latest major to stay up to date

## Changeset


.changeset/thin-cooks-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-config-godaddy": minor
```
jpage-godaddy, line 2: Is there a new feature here, or should this be a patch?

jpage-godaddy (approved):

.changeset/thin-cooks-flow.md
rmarkins-godaddy: yeah thought with the amount of changes a minor made sense, but looking back it's really just a fix that was missed originally

kbader-godaddy (approved):

PR #1130, opened 2025-08-29
chore: add pnpm workspace configuration to manage packages

rxmarbles:
## Summary

Migrate to pnpm to manage packages

## Changeset

- ~[ ] Changeset added ([docs](https://github.com/godaddy/javascript#submit-a-changeset))~

## Test Plan

All CI tests pass


GOVERNANCE.md
```diff
@@ -0,0 +1,66 @@
...
+
+Maintainers have commit permissions. They are responsible for creating release
+commits. They also are responsible for maintaining the Gasket
+[roadmap](./ROADMAP.md). Finally, they are responsible for closing out issues
+that are duplicates, not reproducible or otherwise should not be addressed. New
+maintainers are added by existing maintainers at their discretion. If you want
+to become a maintainer, start by looking at Gasket's backlog of issues. You can
+triage existing issues, provide fixes, and/or create issues that will improve
+Gasket for the community at large. You can view the current list of maintainers
+in the [Gasket Github group].
```
agerard-godaddy, line 20: Clean up gasket stuff in this file, and there is no roadmap.

.nvmrc
```diff
@@ -1 +1 @@
-22
+24
```
mmicco-godaddy, line 1: Preference over this vs putting it in package.json? I tend to use the package json so I can avoid the single odd file?  This is not a suggestion, as either way works.

package.json
```diff
@@ -1,21 +1,24 @@
 {
   "name": "godaddy-style",
-  "version": "6.0.0",
+  "version": "0.0.0",
   "private": true,
   "description": "The approach to JavaScript at GoDaddy. We think it's pretty decent.",
   "type": "module",
   "engines": {
+    "node": ">=22"
```
mmicco-godaddy, line 8: Should this be 24 to match .nvmrc?

package-lock.json
mmicco-godaddy, line 1: Is this even needed anymore when using pnpm?  Just asking because I removed it from some repos I converted (maybe I should have kept it?)

mmicco-godaddy (commented):
Just comments.

package.json
rxmarbles: Not necessarily. This indicates that the package can support Node version greater than 22. If I remove and/or upgrade that it would indicate that it would support only Node@24. Honestly we should position around `devEngines` but `setup-node` doesn't support that field yet. Though there is an open PR to reference it

PR #1355, opened 2026-04-20
fix(react): use pickup location address for tax calculation on billing updates

nlaverde-godaddy:
## Summary

When a customer selects in-person pickup and edits their billing address (even just deleting and re-typing a digit), the tax calculation was using the wrong address, resulting in incorrect tax amounts.

#Root cause
Root Cause: In`use-update-order.ts`, when the billing address was updated, `updateTaxes.mutate()` was called with undefined, causing the backend to use the customer's billing address instead of the pickup location address for tax calculation.

Example of the bug:

Pickup location zipcode: 30143 (7% tax rate)
Customer billing zipcode: 30734 (8% tax rate)
First calculation: Correctly uses pickup location → 7% ✅
After editing billing address: Uses customer address → 8% ❌ (now fixed)
Related ticket: [https://godaddy-corp.atlassian.net/browse/CIEC-11081](https://godaddy-corp.atlassian.net/browse/CIEC-11081)

## Changeset

- [ ] Changeset added ([docs](https://github.com/godaddy/javascript#submit-a-changeset))

## Test Plan

Test Plan
Unit Tests (9 new tests):

✅ Returns pickup location address for PICKUP delivery method
✅ Returns correct location when multiple locations exist
✅ Returns undefined if pickup location not found
✅ Returns undefined if locations array is empty
✅ Returns undefined if locations is undefined
✅ Returns undefined for SHIP delivery method
✅ Returns undefined when delivery method is undefined
✅ Handles pickupLocationId being undefined for pickup
✅ Handles empty string pickupLocationId for pickup

Screenshot 2026-04-20 at 12 19 47

### DIFF
--- a/.changeset/seven-signs-ask.md
+++ b/.changeset/seven-signs-ask.md
@@ -1,0 +1,5 @@
+---
+"@godaddy/react": major
+---
+
+modify use-update-orders.ts to detect pickup orders and always send the pickup location address for tax calculation, ensuring consistent tax rates based on where the customer picks up their order.
--- a/.changeset/seven-signs-ask.md
+++ b/.changeset/seven-signs-ask.md
@@ -1,5 +1,5 @@
 ---
-"@godaddy/react": major
+"@godaddy/react": patch
 ---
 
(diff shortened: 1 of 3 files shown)


pbennett1-godaddy (approved):
great catch!

PR #1379, opened 2026-05-28
resync order shipping of shipping lines and lineItems fulfillment NONE

pbennett1-godaddy:
Summary                                                                                                                                                                                                   
                                                                                                                                                                                                           
 Fixes a checkout edge case where a customer can return to cart after selecting a shipping method, replace or add line items, and then land back on checkout with stale shippingLines but new line items   
 still set to fulfillmentMode: "NONE".                                                                                                                                                                     
                                                                                                                                                                                                           
 Because checkout saw the existing shippingLines, it treated shipping as already selected and allowed checkout to proceed. However, confirmCheckoutSession later removes shipping lines/totals when no     
 line items have fulfillmentMode: "SHIP".                                                                                                                                                                  
                                                                                                                                                                                                           
 This change updates the shipping method application logic so that when checkout detects existing shipping lines but line items still need shipping fulfillment, it reapplies the selected shipping        
 method. Reapplying the shipping method causes checkout-api to recalculate shipping and update line items to fulfillmentMode: "SHIP".                                                                      
                                                                                                                                                                                                           
 Changeset                                                                                                                                                                                                 
                                                                                                                                                                                                           
 - [X] Changeset added (docs (https://github.com/godaddy/javascript#submit-a-changeset))                                                                                                                       
                                                                                                                                                                                                           
 Test Plan                                                                                                                                                                                                 
                                                                                                                                                                                                           
 - Ran React package typecheck:                                                                                                                                                                            
                                                                                                                                                                                                           
 ```bash                                                                                                                                                                                                   
   pnpm --filter @godaddy/react typecheck                                                                                                                                                                  
 ```                                                                                                                                                                                                       
                                                                                                                                                                                                           
 - Ran targeted React tests:                                                                                                                                                                               
                                                                                                                                                                                                           
 ```bash                                                                                                                                                                                                   
   pnpm --filter @godaddy/react test -- shipping-method                                                                                                                                                    
 ```                                                                                                                                                                                                       
                                                                                                                                                                                                           
 - Manually verified the checkout scenario:                                                                                                                                                                
     1. Start checkout with shipping enabled.                                                                                                                                                              
     2. Select/apply a shipping method.                                                                                                                                                                    
     3. Return to cart.                                                                                                                                                                                    
     4. Remove the existing line item and add a new/same item.                                                                                                                                             
     5. Return to checkout.                                                                                                                                                                                
     6. Confirm checkout reapplies the shipping method and updates the new line item from fulfillmentMode: "NONE" to fulfillmentMode: "SHIP".

### DIFF
--- a/packages/react/src/components/checkout/shipping/shipping-method.tsx
+++ b/packages/react/src/components/checkout/shipping/shipping-method.tsx
@@ -1,3 +1,4 @@
+import { useQueryClient } from '@tanstack/react-query';
 import { useEffect, useRef } from 'react';
 import { useFormContext } from 'react-hook-form';
 import { useCheckoutContext } from '@/components/checkout/checkout';
@@ -12,6 +13,10 @@ import { useUpdateTaxes } from '@/components/checkout/order/use-update-taxes';
 import { useIsPaymentDisabled } from '@/components/checkout/payment/utils/use-is-payment-disabled';
 import { ShippingMethodSkeleton } from '@/components/checkout/shipping/shipping-method-skeleton';
 import { filterAndSortShippingMethods } from '@/components/checkout/shipping/utils/filter-shipping-methods';
+import {
+  getShippingFulfillmentSyncKey,
+  shouldApplyShippingMethod,
+} from '@/components/checkout/shipping/utils/should-apply-shipping-method';
 import { useApplyShippingMethod } from '@/components/checkout/shipping/utils/use-apply-shipping-method';
 import { useDraftOrderShippingMethods } from '@/components/checkout/shipping/utils/use-draft-order-shipping-methods';
 import { useFormatCurrency } from '@/components/checkout/utils/format-currency';
@@ -47,6 +52,7 @@ export function ShippingMethodForm() {
   const { t } = useGoDaddyContext();
   const { session, isConfirmingCheckout } = useCheckoutContext();
   const updateTaxes = useUpdateTaxes();
+  const queryClient = useQueryClient();
   const isPaymentDisabled = useIsPaymentDisabled();
 
   const { data: shippingMethodsData, isLoading: isShippingMethodsLoading } =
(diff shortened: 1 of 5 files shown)


packages/react/src/components/checkout/shipping/shipping-method.tsx
```diff
@@ -170,6 +201,7 @@ export function ShippingMethodForm() {
           hadShippingMethods: true,
           wasPickup: false,
           clearedShippingMethod: false,
+          syncingFulfillmentKey: needsMutation ? fulfillmentSyncKey : null,
```
wcole1-godaddy: This records the fulfillment sync key immediately after dispatching `applyShippingMethod.mutate(...)`, before we know the mutation/refetch actually succeeded, and the key is never cleared when the order later shows all line items fulfilled. If the reapply fails or line items are not refreshed, or if a later cart edit reuses the same line item id and resets `fulfillmentMode` to `NONE`, subsequent renders will treat this key as already synced and skip the reapply. That can leave the stale `shippingLines` + `NONE` line item state this PR is trying to fix. Can we clear this key once `hasLineItemsMissingShippingFulfillment` becomes false and/or only store it after a successful mutation/refetch? It may also be worth invalidating/refetching `draft-order` here so `lineItems` are refreshed after the reapply.

packages/react/src/components/checkout/shipping/shipping-method.tsx
```diff
@@ -141,25 +166,53 @@ export function ShippingMethodForm() {
...
+              if (!isFulfillmentSync || !session?.id) return;
+
+              queryClient.invalidateQueries({
+                queryKey: ['draft-order', session.id],
+              });
+            },
+            onError: () => {
+              if (!isFulfillmentSync) return;
+
+              lastProcessedStateRef.current = {
```
wcole1-godaddy: This clears the fulfillment guard on mutation error, but the effect will immediately run again once `applyShippingMethod.isPending` flips back to false while the same line items still have the same `fulfillmentSyncKey`. Because `shouldApplyShippingMethod` then sees `inFlightFulfillmentKey: null`, it will dispatch the same mutation again; repeated failures can create an automatic retry loop and hammer the shipping API. Can we avoid retrying from the effect after an error (for example by tracking a failed key / surfacing the error, and only retrying when the cart or selected method changes)?

wcole1-godaddy (approved):
Approved. Latest revision addresses my previous concerns, and CI is green.

PR #1439, opened 2026-08-24
VNEXT-83184: Redesign checkout coupon UI

abansal2-godaddy:
## Summary

Implements [VNEXT-83184](https://godaddy-corp.atlassian.net/browse/VNEXT-83184): redesign the checkout **coupon / promotion code** UI (default → filled → success → error).

**Where this lives:** `@godaddy/react` `DiscountStandalone` (used by Bruce when `enablePromotionCodes` is on).

## Design

1. Combined bordered row (`h-12`, theme tokens) — input + Apply (or clear on error)
2. Compact helper text under the field for errors
3. Full-width applied bar — check + code + amount + remove (`accent` surface, `foreground` text)
4. **One coupon at a time** — entry field hidden while a code is applied
5. **Coupon heading is host-owned** via `checkout.summary.discount.before` / `.after` (not shipped in the package)

## Error copy

Maps GraphQL codes via `apiErrors`, falls back to `failedToApply` (does not claim every failure is an “invalid” code).

## What changed

- `discount-standalone.tsx` / `discount-applied-bar.tsx`
- Targets: `checkout.summary.discount.before|after`
- Existing locale keys only (`placeholder`, `enterCode`, `apply`, …) — no new en-US-only strings
- Unit tests + changeset

## Test plan

- [x] Discount unit tests pass locally
- [ ] After publish + Bruce bump: default / filled / success / error on checkout
- [ ] Host injects heading via `checkout.summary.discount.before` if desired
- [ ] Applied bar readable under default theme; remove restores input


packages/react/src/components/checkout/discount/discount-standalone.tsx
```diff
@@ -1,111 +1,141 @@
...
 import { useDraftOrder } from '@/components/checkout/order/use-draft-order';
 import { useIsPaymentDisabled } from '@/components/checkout/payment/utils/use-is-payment-disabled';
 import { useGoDaddyContext } from '@/godaddy-provider';
 import { GraphQLErrorWithCodes } from '@/lib/graphql-with-errors';
+import { cn } from '@/lib/utils';
+import type { DraftOrder } from '@/types';
 import { eventIds } from '@/tracking/events';
 import { TrackingEventType, track } from '@/tracking/track';
-import { Discounts } from './discounts';
 import type { DiscountFormProps } from './types';
```
abansal2-godaddy, line 16: **Why these imports changed / why the old ones are gone from this file**

Previously this file composed:
- `DiscountInput`
- `DiscountApplyButton`
- `DiscountErrorList`
- `Discounts` / `DiscountTag`

Those matched the **old** UX (separate input + button, large alert errors, chip tags with code only).

VNEXT-83184 needs a **different** UX, so those imports were removed **from this path**:
- Input + Apply (or clear) must live in **one** bordered row → not standalone `DiscountInput` / `DiscountApplyButton`
- Errors are a **short helper** under the field → not `DiscountErrorList`
- Success needs **code + formatted amount + remove** in a full-width bar → not chip `DiscountTag`s

The old modules are still in the package/exported for now; they just are not used by `DiscountStandalone` anymore. Cleanup can be a follow-up.

packages/react/src/components/checkout/discount/discount-standalone.tsx
```diff
@@ -215,38 +233,97 @@ export function DiscountStandalone({
...
+      <div className='flex flex-col gap-1.5'>
+        <div
+          className={cn(
+            'flex h-14 items-center justify-between rounded-md border bg-white py-2 pl-4 pr-2',
+            hasError
+              ? 'border-[#EF4444]'
+              : isFocused || hasInputValue
+                ? 'border-[#2563EB]'
+                : 'border-[#D1D5DB]'
+          )}
```
abansal2-godaddy, line 308: **Why the layout is inlined here**

This combined bordered row (input + Apply, border color by default/filled/error) is the core of the redesign. Wiring the old separate input/button components into this shell would mean rewriting their APIs anyway, so the field markup lives here and keeps the four visual states in one place.

packages/react/src/components/checkout/discount/discount-applied-bar.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export function DiscountAppliedBar({
+  code,
+  amount,
+  currencyCode,
+  inputInMinorUnits = true,
+  onRemove,
+  isRemoving,
+}: DiscountAppliedBarProps) {
```
abansal2-godaddy, line 24: **Why a new `DiscountAppliedBar` instead of tweaking `DiscountTag`**

`DiscountTag` is a small chip (tag icon + code + X) and has **no amount**. The success state in the design is a full-width green bar: check + code + `– $amount` + remove.

That is a different component shape, so we added this bar for `DiscountStandalone` rather than overloading the chip. `DiscountTag` / `Discounts` remain in the package unused by this path until a cleanup PR.

abansal2-godaddy (commented):
Reviewer notes (also in the PR description): why the layout changed, and why the old discount imports were dropped from  — without leaving that rationale in the source files.

abansal2-godaddy:
## Reviewer note: layout vs old imports

Details are on the **Files changed** tab (inline comments) and in the updated PR description.

**Short version**
- New design is a combined input+Apply row, compact error helper, and a success bar with **amount** — not the old separate input/button/alert/chip composition.
- That is why `DiscountStandalone` dropped `DiscountInput` / `DiscountApplyButton` / `DiscountErrorList` / `Discounts` / `DiscountTag` and why `DiscountAppliedBar` was added.
- Rationale is documented in the PR only (not as source comments).

pbennett1-godaddy:
@abansal2-godaddy - is the main goal here to show each coupon codes value instead of a single discount total? I'd prefer not to make huge design changes if possible. We currently follow a similar pattern to other checkouts so I don't think it would be confusing users thinking it's mandatory.

abansal2-godaddy:
> @abansal2-godaddy - is the main goal here to show each coupon codes value instead of a single discount total? I'd prefer not to make huge design changes if possible. We currently follow a similar pattern to other checkouts so I don't think it would be confusing users thinking it's mandatory.

No @pbennett1-godaddy ,It's more of a UI and experience change only with proper error states and intent when a C1 applies a coupon.
Also how can I test these changes on bruce storefront on dev/test environment?  
Screenshot 2026-08-27 at 11 33 59 AM

pbennett1-godaddy:
> No @pbennett1-godaddy ,It's more of a UI and experience change only with proper error states and intent when a C1 applies a coupon. Also how can I test these changes on bruce storefront on dev/test environment? 

you can use the nextjs sample app in this repo to test against a dev/test env store. Can you look into what discount GPAs actually return on failure? I don't think we should have a default of "Coupon isn't valid" cause I think there is basically only a single error of "coupon code failed to apply" from the BE. Invalid would lead customers to believe the coupon they tried to use even if there was an error was an invalid code.

abansal2-godaddy:
> > No @pbennett1-godaddy ,It's more of a UI and experience change only with proper error states and intent when a C1 applies a coupon. Also how can I test these changes on bruce storefront on dev/test environment?
> 
> you can use the nextjs sample app in this repo to test against a dev/test env store. Can you look into what discount GPAs actually return on failure? I don't think we should have a default of "Coupon isn't valid" cause I think there is basically only a single error of "coupon code failed to apply" from the BE. Invalid would lead customers to believe the coupon they tried to use even if there was an error was an invalid code.

Yes you are right, I'll check what we get as response under various cases from BE and have the error state render the intended failure

PR #1451, opened 2026-09-09
Add Razorpay as a standard checkout payment method.

pchawla-godaddy:
Wire the C2 checkout UI to authorize, open Razorpay Standard Checkout, encode the signed callback, and confirm, matching the checkout-api contract.

## Changeset

- [ ] Changeset added ([docs](https://github.com/godaddy/javascript#submit-a-changeset))

## Test Plan

https://github.com/user-attachments/assets/eb337a18-be9f-4d6a-ac74-604506cd6167

### DIFF
--- a/.changeset/bright-cats-pay.md
+++ b/.changeset/bright-cats-pay.md
@@ -1,0 +1,6 @@
+---
+'@godaddy/react': minor
+'@godaddy/localizations': patch
+---
+
+Add Razorpay as a standard checkout payment method.
--- a/packages/react/src/components/checkout/checkout.tsx
+++ b/packages/react/src/components/checkout/checkout.tsx
@@ -94,6 +94,10 @@ export type CCAvenueConfig = {
   accessCodeId: string;
 };
 
+export type RazorpayConfig = {
+  publicToken: string;
+};
+
 interface CheckoutContextValue {
   elements?: CheckoutElements;
   targets?: Partial<
@@ -108,6 +112,7 @@ interface CheckoutContextValue {
   paypalConfig?: PayPalConfig;
   mercadoPagoConfig?: MercadoPagoConfig;
   ccavenueConfig?: CCAvenueConfig;
+  razorpayConfig?: RazorpayConfig;
   isConfirmingCheckout: boolean;
   setIsConfirmingCheckout: (isConfirming: boolean) => void;
   checkoutErrors?: string[] | undefined;
@@ -214,6 +219,7 @@ export interface CheckoutProps {
   paypalConfig?: PayPalConfig;
   mercadoPagoConfig?: MercadoPagoConfig;
   ccavenueConfig?: CCAvenueConfig;
+  razorpayConfig?: RazorpayConfig;
   layout?: LayoutSection[];
   direction?: 'ltr' | 'rtl';
   showStoreHours?: boolean;
@@ -239,6 +245,7 @@ export function Checkout(props: CheckoutProps) {
     paypalConfig,
     mercadoPagoConfig,
     ccavenueConfig,
+    razorpayConfig,
     isCheckoutDisabled,
   } = props;
 
@@ -333,6 +340,7 @@ export function Checkout(props: CheckoutProps) {
           mercadoPagoConfig,
           paypalConfig,
           ccavenueConfig,
+          razorpayConfig,
           requiredFields,
           isConfirmingCheckout,
           setIsConfirmingCheckout,
--- a/packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
+++ b/packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
@@ -1,0 +1,261 @@
+import { LoaderCircle } from 'lucide-react';
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { useFormContext } from 'react-hook-form';
+import { useCheckoutContext } from '@/components/checkout/checkout';
+import { useAuthorizeCheckout } from '@/components/checkout/payment/utils/use-authorize-checkout';
+import { encodeRazorpayPaymentToken } from '@/components/checkout/payment/utils/razorpay-payment-token';
+import {
+  PaymentProvider,
+  useConfirmCheckout,
+} from '@/components/checkout/payment/utils/use-confirm-checkout';
+import { useFlushCheckoutSync } from '@/components/checkout/payment/utils/use-flush-checkout-sync';
+import { useIsPaymentDisabled } from '@/components/checkout/payment/utils/use-is-payment-disabled';
+import { useLoadRazorpay } from '@/components/checkout/payment/utils/use-load-razorpay';
+import { normalizePhoneForRazorpay } from '@/components/checkout/utils/checkout-transformers';
+import { Button } from '@/components/ui/button';
+import { useGoDaddyContext } from '@/godaddy-provider';
+import { GraphQLErrorWithCodes } from '@/lib/graphql-with-errors';
+import { PaymentMethodType } from '@/types';
+
+export const RAZORPAY_CALLBACK_TIMEOUT_MS = 2 * 60 * 1000;
+
+type RazorpaySuccessResponse = {
+  razorpay_payment_id?: string;
+  razorpay_order_id?: string;
+  razorpay_signature?: string;
+};
+
+type RazorpayOptions = {
+  key: string;
+  amount: number;
+  currency: string;
+  name?: string;
+  description: string;
+  order_id: string;
+  prefill?: {
+    name?: string;
+    email?: string;
+    contact?: string;
+  };
+  handler: (response: RazorpaySuccessResponse) => void;
+  modal: {
+    ondismiss: () => void;
+  };
+};
+
+type RazorpayInstance = {
+  open: () => void;
+  close: () => void;
+  on: (
+    event: 'payment.failed',
+    handler: (response: unknown) => void
+  ) => void;
+};
+
+type RazorpayConstructor = new (options: RazorpayOptions) => RazorpayInstance;
+
+function getRazorpayConstructor(): RazorpayConstructor | undefined {
+  return (window as Window & { Razorpay?: RazorpayConstructor }).Razorpay;
+}
+
+export function RazorpayCheckoutButton() {
+  const { t } = useGoDaddyContext();
+  const {
+    razorpayConfig,
+    session,
+    setCheckoutErrors,
+    isConfirmingCheckout,
+  } = useCheckoutContext();
+  const form = useFormContext();
+  const authorizeCheckout = useAuthorizeCheckout();
+  const confirmCheckout = useConfirmCheckout();
+  const flushCheckoutSync = useFlushCheckoutSync();
+  const isPaymentDisabled = useIsPaymentDisabled();
+  const { isRazorpayLoaded, isRazorpayLoadFailed } = useLoadRazorpay();
+  const [isWidgetOpen, setIsWidgetOpen] = useState(false);
+  const [error, setError] = useState('');
+  const callbackTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  const clearCallbackTimeout = useCallback(() => {
+    if (callbackTimeoutRef.current) {
+      clearTimeout(callbackTimeoutRef.current);
+      callbackTimeoutRef.current = null;
+    }
+  }, []);
+
+  useEffect(() => clearCallbackTimeout, [clearCallbackTimeout]);
+
+  const handlePaymentSuccess = useCallback(
+    async (response: RazorpaySuccessResponse) => {
+      clearCallbackTimeout();
+      const paymentId = response.razorpay_payment_id;
+      const orderId = response.razorpay_order_id;
+      const signature = response.razorpay_signature;
+      if (!paymentId || !orderId || !signature) {
+        setError(t.errors.errorProcessingPayment);
+        setIsWidgetOpen(false);
+        return;
+      }
+
+      try {
+        const paymentToken = encodeRazorpayPaymentToken({
+          v: 1,
+          paymentId,
+          orderId,
+          signature,
+        });
+        await confirmCheckout.mutateAsync({
+          paymentToken,
+          paymentType: PaymentMethodType.RAZORPAY,
+          paymentProvider: PaymentProvider.RAZORPAY,
+        });
+        setError('');
+      } catch (err: unknown) {
+        if (err instanceof GraphQLErrorWithCodes) {
+          setCheckoutErrors(err.codes);
+        } else {
+          setError(t.errors.errorProcessingPayment);
+        }
+      } finally {
+        setIsWidgetOpen(false);
+      }
+    },
+    [
+      confirmCheckout,
+      clearCallbackTimeout,
+      setCheckoutErrors,
+      t.errors.errorProcessingPayment,
+    ]
+  );
+
+  const handlePaymentFailure = useCallback(() => {
+    setError(t.errors.errorProcessingPayment);
+  }, [t.errors.errorProcessingPayment]);
+
+  const handleClick = async () => {
+    if (isWidgetOpen || authorizeCheckout.isPending || isConfirmingCheckout) {
+      return;
+    }
+
+    const valid = await form.trigger();
+    if (!valid) {
+      const firstError = Object.keys(form.formState.errors)[0];
+      if (firstError) form.setFocus(firstError);
+      return;
+    }
+
+    setCheckoutErrors(undefined);
+    setError('');
+
+    try {
+      const { latestOrder } = await flushCheckoutSync({
+        includeCurrentFormDiff: true,
+      });
+      const total = latestOrder?.totals?.total;
+      if (
+        !latestOrder?.id ||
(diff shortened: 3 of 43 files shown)


packages/react/src/components/checkout/checkout.tsx
```diff
@@ -94,6 +94,10 @@ export type CCAvenueConfig = {
   accessCodeId: string;
 };
 
+export type RazorpayConfig = {
+  publicToken: string;
```
rsharma2-godaddy, line 98: @pchawla-godaddy 
the public token no longer comes from the config, this was updated and comes from the `paymentReferences` please refer the PR for checkout-api
https://github.com/gdcorp-commerce/checkout-api/pull/196

rsharma2-godaddy (commented):
we might have ro re-attempt this PR, since some changes were requested
can you also add the button to pay with razorpay with the icon in the video

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
```diff
@@ -0,0 +1,261 @@
...
+          paymentId,
+          orderId,
+          signature,
+        });
+        await confirmCheckout.mutateAsync({
+          paymentToken,
+          paymentType: PaymentMethodType.RAZORPAY,
+          paymentProvider: PaymentProvider.RAZORPAY,
+        });
+        setError('');
```
abarala-godaddy, line 112: why we need here to set empty error?

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+    const valid = await form.trigger();
+    if (!valid) {
+      const firstError = Object.keys(form.formState.errors)[0];
+      if (firstError) form.setFocus(firstError);
+      return;
+    }
+
+    setCheckoutErrors(undefined);
+    setError('');
```
abarala-godaddy, line 148: why we need setCheckoutErrors and setError with empty and undefined value

packages/react/src/components/checkout/checkout.tsx
pchawla-godaddy: Addressed in a5d9736. The frontend no longer accepts `razorpayConfig` or reads the public token from client configuration. After `authorizeCheckoutSession`, it now reads the Razorpay order ID from `fundingSource.paymentReference` and the public key from the `references` entry whose type is `MERCHANT_PUBLIC_KEY`, matching checkout-api PR #196.

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
pchawla-godaddy: Addressed in a5d9736. I removed the redundant `setError("")` after successful confirmation; the success path redirects, so clearing local error state there was unnecessary.

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
pchawla-godaddy: These resets are intentionally kept at the start of a new payment attempt. `setCheckoutErrors(undefined)` clears any prior GraphQL error codes, while `setError("")` clears a prior local SDK/authorization/payment error. Without them, a corrected retry can continue showing the stale error from the previous attempt. The redundant reset after successful confirmation has been removed.

.changeset/bright-cats-pay.md
```diff
@@ -0,0 +1,6 @@
+---
+'@godaddy/react': minor
```
pbennett1-godaddy:
```suggestion
'@godaddy/react': patch
```

packages/react/src/components/checkout/payment/utils/use-get-selected-payment-method.ts
```diff
@@ -10,14 +10,16 @@ export function useGetSelectedPaymentMethod(
...
     if (!paymentMethod || !session?.paymentMethods) return null;
 
-    const methodConfig = session.paymentMethods[
-      paymentMethod as PaymentMethodValue
-    ] as PaymentMethodConfig;
+    const paymentMethods = session.paymentMethods as unknown as Partial<
+      Record<PaymentMethodValue, PaymentMethodConfig>
+    >;
+    const methodConfig = paymentMethods[paymentMethod];
+    if (!methodConfig) return null;
```
rsharma2-godaddy, line 19: have we done `null` handing everywhere, since we are changing the contract

packages/react/src/components/checkout/payment/utils/use-load-razorpay.ts
```diff
@@ -0,0 +1,90 @@
+import { useEffect, useState } from 'react';
+
+const RAZORPAY_SDK_ID = 'razorpay-sdk';
+const RAZORPAY_SDK_URL = 'https://checkout.razorpay.com/v1/checkout.js';
+
+let isRazorpayLoaded = false;
+let isRazorpayScriptRequested = false;
```
rsharma2-godaddy: can we not use global variables, it could be risky

packages/react/src/components/checkout/payment/utils/use-load-razorpay.ts
```diff
@@ -0,0 +1,90 @@
+import { useEffect, useState } from 'react';
```
rsharma2-godaddy: there is a standard pattern to load scripts in react using `useSyncExternalStore`

rsharma2-godaddy: we can also have a retry incorporated in this flow, essentially things are stuck if script execution fails

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+    try {
+      const { latestOrder } = await flushCheckoutSync({
```
rsharma2-godaddy, line 157: @pchawla-godaddy could you check the implmentation for `flushCheckoutSync` in detail

packages/react/src/components/checkout/payment/utils/use-get-selected-payment-method.ts
pchawla-godaddy: Yes — I audited every `session.paymentMethods` access. The production sites already guard with optional chaining or truthiness, so a null container or a configured key whose value is null does not crash.

I still made the local record casts explicitly include `null` at both the container and per-method level, and added coverage for `paymentMethods: null` and `{ razorpay: null }`.

Also added `razorpay { processor checkoutTypes }` to `CreateCheckoutSessionMutation` (it was previously only on `GetCheckoutSessionQuery`) so a newly created session does not omit Razorpay. Landed in bf415b9.

packages/react/src/components/checkout/payment/utils/use-load-razorpay.ts
pchawla-godaddy: Agreed — the module-level `isRazorpayLoaded` / `isRazorpayScriptRequested` flags are gone.

Load state now lives in `RazorpayLoaderProvider`, and the DOM script tag is the source of truth for whether a load is already in flight. Applied the provider-based loader from your patch in bf415b9.

packages/react/src/components/checkout/payment/utils/use-load-razorpay.ts
pchawla-godaddy: Went with the provider-based loader from your later patch rather than `useSyncExternalStore`, so load state is React-owned and shared by consumers under `RazorpayLoaderProvider`.

`RazorpayCheckoutButton` wraps the inner button in that provider. Landed in bf415b9.

pchawla-godaddy: Retries are in the new loader. A failed script load retries up to 3 times with 1s / 2s / 3s backoff. After those are exhausted we set `isRazorpayLoadFailed` so the button can surface the error instead of staying stuck.

A `load` event without `window.Razorpay` is also treated as a failure. Covered in `razorpay-loader-context.test.tsx`. Landed in bf415b9.

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
pchawla-godaddy: Checked `flushCheckoutSync` in detail. `includeCurrentFormDiff: true` is required here: it drains debounced buyer edits, rebuilds current form values, refetches after a patch, and returns `latestOrder`, which we use for Razorpay amount and prefill.

`useAuthorizeCheckout` still does a second basic flush internally. After this call that should be a no-op, so I left the shared hook unchanged.

Added a test that holds the flush promise open and asserts authorize does not start until sync settles. Landed in bf415b9.

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
```diff
@@ -0,0 +1,263 @@
...
+      });
+      widget.on('payment.failed', handlePaymentFailure);
+
+      setIsWidgetOpen(true);
+      callbackTimeoutRef.current = setTimeout(() => {
+        callbackTimeoutRef.current = null;
+        widget.close();
+        setIsWidgetOpen(false);
+        setError(t.errors.errorProcessingPayment);
+      }, RAZORPAY_CALLBACK_TIMEOUT_MS);
```
wcole1-godaddy: [P1] Avoid interrupting active payments after two minutes

This timer starts when the widget opens and unconditionally closes it after two minutes, including while a shopper is entering details or completing bank authentication. Razorpay documents transaction timeouts of 3–15 minutes after a payment attempt begins, so this can cut off a valid payment before its success callback reaches checkout. Please remove the unconditional cutoff and let the SDK manage active payment expiry.

Reference: https://razorpay.com/docs/payments/payment-gateway/web-integration/standard/troubleshooting-faqs/

packages/react/src/components/checkout/payment/utils/razorpay-loader-context.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    function loadScript() {
+      // The DOM is the source of truth for whether a script tag is already
+      // in flight, so no module-level flag is needed to dedupe requests.
+      const existingScript = document.getElementById(
+        RAZORPAY_SDK_ID
+      ) as HTMLScriptElement | null;
+      if (existingScript) {
+        attachListeners(existingScript);
+        return;
```
wcole1-godaddy: [P2] Recover failed SDK script tags when the provider remounts

An existing tag is not necessarily still loading. If the shopper switches payment methods and the SDK request fails while this provider is unmounted, cleanup has detached the error listener but left the tag in the DOM. Returning to Razorpay attaches listeners to that already-failed tag; no new event arrives, so both load flags remain false and payment stays disabled without an error. A load event without a constructor followed by remount has the same problem. I reproduced both paths with a source-level probe. Please preserve the script completion/failure status independently of provider mounts and replace failed tags instead of waiting on them again.

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const handlePaymentFailure = useCallback(() => {
+    setError(t.errors.errorProcessingPayment);
+  }, [t.errors.errorProcessingPayment]);
+
+  const handleClick = async () => {
+    if (isWidgetOpen || authorizeCheckout.isPending || isConfirmingCheckout) {
+      return;
+    }
+
+    const valid = await form.trigger();
```
wcole1-godaddy: [P2] Lock the payment attempt before awaiting validation

The busy guard runs before asynchronous validation/synchronization, but no attempt lock is acquired. If two clicks enter before validation resolves, both pass the guard and later authorize and open their own widgets; pending state set by the first authorization cannot stop the second handler that already passed this check. A source-level probe holding validation open produced two authorizations and two widget opens. Please acquire a synchronous ref lock before the first await, expose corresponding busy state, and release it on validation failure, dismissal, or completion. Add coverage for two clicks while validation or synchronization is pending.

wcole1-godaddy (changes requested):
Requesting changes for three payment lifecycle issues: the unconditional two-minute widget cutoff, SDK load recovery after remount, and overlapping payment attempts. Details and suggested fixes are inline.

Validation: CI passes. Isolated source-level probes reproduced the stale-script state and duplicate authorization/widget opens, and verified that the timer closes the widget without a payment attempt. Vitest was not run locally because dependencies are not installed; no live gateway transaction was performed.

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
pchawla-godaddy: Addressed in cd64112. Removed the unconditional two-minute timer and the client-side `widget.close()` path. Razorpay now owns payment expiry (including its documented 3–15 minute transaction timeout), while `modal.ondismiss` still releases the local payment-attempt state when the shopper closes Checkout.

The old timeout test was replaced with coverage asserting that opening Checkout does not schedule a two-minute cutoff.

packages/react/src/components/checkout/payment/utils/razorpay-loader-context.tsx
pchawla-godaddy: Addressed in cd64112. The script element now persists its outcome in `data-status` (`loading`, `loaded`, or `failed`) using status listeners that intentionally survive provider unmounts.

On remount, an in-flight `loading` tag is reused, while a failed, completed-without-constructor, or untracked tag is removed and replaced. Added tests for remount after both an unobserved script error and a load event without `window.Razorpay`.

packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx
pchawla-godaddy: Addressed in cd64112. `attemptLockRef` is now acquired synchronously before `form.trigger()` and checked before any await. `isAttempting` exposes the lock through the busy/disabled UI, and the lock is released on invalid validation, setup/authorization failure, modal dismissal, or confirmation completion.

Added coverage that holds validation pending, clicks Pay twice, and verifies one validation, one sync, one authorization, and one widget open.

wcole1-godaddy:
## Critic review

Reviewed commit `06be7664647524199422455bb294277f68ca8e31`. Two actionable lifecycle findings:

### [P2] Preserve confirmation errors across the parent's unmount

`packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx:110–115`

After Razorpay reports success, `useConfirmCheckout` sets `isConfirmingCheckout=true`, causing `PaymentForm` to replace the Razorpay button with its generic processing button. If confirmation then rejects with a transport error, the catch calls `setError` on the unmounted component. Once confirmation state resets, the newly mounted button has no error, so the shopper receives no indication that confirmation failed.

**Suggested fix:** Store the failure in surviving checkout-level state, or keep the Razorpay component mounted during confirmation. Add coverage for a confirmation rejection while the parent transitions into and out of its processing state.

### [P2] Invalidate pending widget initialization when the selected method unmounts

`packages/react/src/components/checkout/payment/checkout-buttons/razorpay/razorpay.tsx:216–217`

The payment-method accordion remains enabled while Razorpay authorization is pending. If the shopper switches methods before authorization resolves, the old component's async continuation still constructs and opens the Razorpay widget after unmount, over the newly selected payment flow. Switching away and back also creates a fresh attempt lock while the original continuation remains active.

**Suggested fix:** Invalidate pending attempts on unmount and check attempt identity after asynchronous boundaries, before opening the SDK. Add coverage for switching methods while authorization is pending.

### Validation

- Both behaviors reproduced in isolated lifecycle tests against the reviewed component.
- CI checks and `@godaddy/react` typecheck pass.
- All 61 tests across the eight focused PR test files pass.
- Full React suite: 818 passed and one billing test timed out at 15 seconds. All 13 tests in that billing file pass on an isolated single-worker rerun.

Existing tests miss these lifecycle transitions: the Razorpay unit tests keep `isConfirmingCheckout` false, while checkout integration tests replace the Razorpay button with an inert mock.

pchawla-godaddy:
Addressed both lifecycle findings in `662a2a8`.

- Confirmation failures now use checkout-level `checkoutErrors`; non-GraphQL/transport failures set `TRANSACTION_PROCESSING_FAILED`, so the error survives the Razorpay button being replaced during confirmation.
- Each payment attempt now has an identity that is invalidated on unmount and checked after validation, draft-order sync, and authorization. Stale continuations cannot construct/open Razorpay after switching methods, and a remounted attempt remains isolated.
- Added lifecycle coverage for confirmation rejection across unmount/remount, authorization resolving after unmount, and switch-away/switch-back with stale authorization.

Validation:
- Razorpay button tests: 10 passed
- Eight focused PR test files: 64 passed
- Full React suite: 80 files / 822 tests passed (`TZ=UTC`)
- `@godaddy/react` typecheck: passed
- `@godaddy/react` lint: passed

The critic findings were a top-level PR comment rather than resolvable review threads. Could you please re-review the latest commit?

PR #1463, opened 2026-09-14
chore(deps): consolidate compatible Dependabot updates

pbennett1-godaddy:
## Summary

Consolidate compatible Dependabot updates while deferring the upgrades that would narrow the requirements of published packages. Retain JSDoc 50.8.0, Mocha 10.5.0, and resolvers 4.1.3; preserve the existing ESLint and React Hook Form peer ranges and omit new engine restrictions from the published packages.

| Dependency | Result | Dependabot PRs superseded after merge |
| --- | --- | --- |
| Next.js | 16.3.4 | #1462, #1454, #1452 |
| Vitest and coverage | Pinned together at 5.0.0 in React and app-connect; grouped for future updates | #1461, #1453 |
| ULID | 3.0.2 | #1460 |
| @changesets/cli | 3.0.3 at the root; remove app-connect's duplicate CLI and release entry point | #1459 |
| @babel/eslint-parser | Remove the unused dependency, Babel core peer, and stale parser option | #1456 |
| globals | 17.12.0; preserve AudioWorklet and disposable-stack globals moved out of the browser set | #1447 |
| changesets/action | v2 (currently 2.1.2) | #1433 |

Deferred major upgrades, not superseded by this PR:

- #1457: eslint-plugin-jsdoc 64, which introduces newer Node requirements.
- #1458: resolvers 5, which requires a newer React Hook Form peer range.
- #1431: eslint-plugin-mocha 12, which removes rule names and synchronous CommonJS compatibility. This PR is already closed.

Release and development tooling:

- Preserve main's manual snapshot workflow and keep the normal Changesets action restricted to pushes. Pass snapshot tags through quoted environment variables.
- Centralize versioning in `scripts/version.js`: empty normal/snapshot requests are successful no-ops, snapshot publishing requires successful versioning of a nonempty changeset, and other versioning errors remain failures. Preserve calculated snapshot versions and use the v4 configuration schema.
- Disable GitHub Release creation, tag pushes, and CLI git-tag creation explicitly.
- The root release command runs Changesets without unconditional builds. Package publish hooks build as needed; React's hook includes its workspace dependencies for fresh checkouts.
- The private root workspace requires Node `^22.13.0 || ^24 || >=26` for its updated tooling; CI remains on Node 24. Published package requirements are unchanged, except removal of the unused Babel peer.

## Changeset

No new changeset included, as requested. The consumer-breaking upgrades and peer/engine restrictions have been removed from this PR.

## Test Plan

- Frozen install passed with pnpm 10.14.0 and Node 24.19.0.
- `pnpm test` passed: seven release-script regression tests, all builds including Next.js, all four ESLint configuration package checks, 593 React tests, and 25 app-connect tests.
- React and localizations typechecks passed.
- Re-ran all ESLint configuration checks after preserving the moved globals.
- Verified the published peer/engine contracts match main, except the removed unused Babel peer.
- Verified synchronous CommonJS loading of the base ESLint configuration and preservation of AudioWorklet/disposable-stack globals.
- App-connect and focused React coverage runs passed during the review fixes.
- Verified release event guards, snapshot output gating, quoted tags, disabled release/tag creation, and read-only Changesets v3.0.3 publish planning.

Live npm publishing and GitHub release creation have not been exercised.


packages/eslint-config-godaddy-react/package.json
```diff
@@ -27,19 +27,19 @@
     "a11y"
   ],
   "peerDependencies": {
-    "@babel/core": ">=7",
-    "eslint": "^10"
+    "@babel/core": "^8.0.0",
```
wcole1-godaddy: **Babel 8 peer for code that never runs.** `index.js` has not imported `@babel/eslint-parser` since the ESLint 9 migration (16301df); the only remnant is a stale `parserOptions.requireConfigFile: false`, which espree ignores. Tightening this peer from `>=7` to `^8.0.0` means any consumer with Babel 7 in their tree (Next, Jest, Storybook toolchains) hits ERESOLVE on install for a parser that is never loaded.

Suggest deleting the `@babel/core` peer, the `@babel/eslint-parser` dependency (line 34), and the `requireConfigFile` line, instead of bumping them. That also removes a second full Babel toolchain from the lockfile.

packages/eslint-config-godaddy/package.json
```diff
@@ -23,18 +23,21 @@
...
+    "eslint": "^10.2.0"
   },
   "dependencies": {
     "@eslint/js": "^10.0.0",
-    "eslint-plugin-jsdoc": "^50.6.9",
+    "eslint-plugin-jsdoc": "^64.3.9",
     "eslint-plugin-json": "^4.0.1",
-    "eslint-plugin-mocha": "^10.5.0",
-    "globals": "^16.0.0"
+    "eslint-plugin-mocha": "^12.0.2",
```
wcole1-godaddy: **eslint-plugin-mocha 12 is ESM-only and drops 11 rules.** Verified live on this head: a CommonJS `eslint.config.cjs` doing `require('eslint-config-godaddy')` now fails with `ERR_REQUIRE_ASYNC_MODULE` on Node 22/24 (v12 has a top-level `await` in `plugin.js`). Consumers overriding any of `mocha/no-setup-in-describe`, `no-skipped-tests`, `no-sibling-hooks`, `no-hooks-for-single-case`, `no-return-and-callback`, `no-async-describe`, `no-global-tests`, `no-top-level-hooks`, `no-empty-description`, `valid-test-description`, `valid-suite-description` get a fatal "Could not find rule in plugin mocha".

The repo only uses `mocha/no-exclusive-tests`, so in-repo tests stay green. This needs at least a changeset and a note; arguably a major for this package and its three wrappers.

packages/eslint-config-godaddy/package.json
```diff
@@ -23,18 +23,21 @@
     "style-guide",
     "eslint"
   ],
+  "engines": {
+    "node": "^22.22.2 || >=24.15.0"
+  },
   "peerDependencies": {
-    "eslint": "^10"
+    "eslint": "^10.2.0"
```
wcole1-godaddy: **No changeset for consumer-facing contract changes.** This PR narrows the `eslint` peer (`^10` -> `^10.2.0`) in four configs, adds `engines`, majors `eslint-plugin-jsdoc`/`mocha`/`globals`, tightens `@babel/core` to ^8 in -react, and raises the `react-hook-form` floor in @godaddy/react, all with no `.changeset/*.md` (changeset-bot flagged this too).

Under changesets nothing publishes until an unrelated changeset bumps these packages, at which point e.g. a `patch` for eslint-config-godaddy ships 9.0.1 that rejects eslint 10.0/10.1 and Node <22.22.2, with no CHANGELOG line. Repo precedent (f47da10, 0fabf67) and CONTRIBUTING.md both call for changesets on peer changes.

package.json
```diff
@@ -14,6 +14,7 @@
     "lint:fix": "pnpm run -r lint:fix",
     "test": "pnpm run build && pnpm run -r --parallel test",
     "version": "changeset version && pnpm install --prefer-offline",
+    "release": "pnpm --filter './packages/*' -r build && changeset publish",
```
wcole1-godaddy: **Pre-build is unconditional and duplicated.** `changeset publish` (v3, default non-artifact mode) runs `pnpm publish` in each package dir, which already fires `prepublishOnly: pnpm run build` in app-connect, localizations and react. So every published package is built twice, including react's three `pnpm dlx @tailwindcss/cli` spawns.

changesets/action@v2 also runs `publish-script` on every push to main with no pending changesets, so a docs-only merge now runs tsdown x3 plus network-dependent dlx before changesets reports "No unpublished projects to publish". Any transient registry failure turns the Release job red with nothing to publish. The old `pnpm publish -r` flow built nothing in that case.

Ordering is not a reason to keep this: v3 publishes in package-graph chunks, so localizations' `prepublishOnly` completes before react's. Suggest `"release": "changeset publish"`, which is what AGENTS.md line 27 already documents.

packages/eslint-config-godaddy/package.json
```diff
@@ -23,18 +23,21 @@
     "style-guide",
     "eslint"
   ],
+  "engines": {
+    "node": "^22.22.2 || >=24.15.0"
```
wcole1-godaddy: **Engines hand-copied with two different ranges.** This is `^22.22.2 || >=24.15.0` (byte-identical to eslint-plugin-jsdoc@64), while root `package.json` line 8 is `^22.22.2 || ^24.15.0 || >=26`, which excludes all of Node 25.x. The three configs that depend on this via `workspace:^` declare no engines at all, `.nvmrc` says 24, and AGENTS.md lines 8/105/358 still say `engines >=22`.

eslint-config-godaddy@9.0.0 on npm has no engines field, so the next publish adds one and `engine-strict` consumers on 22.13-22.22.1 or 24.0-24.14 hard-fail install where 9.0.0 worked. Pick one string, use it in both places (and ideally on the three sibling configs), and update AGENTS.md.

.github/workflows/release.yaml
```diff
@@ -34,9 +34,8 @@ jobs:
 
       - name: Create Release Pull Request or Publish to npm
         id: changesets
-        uses: changesets/action@v1 # v1.7.0
+        uses: changesets/action@v2 # v2.1.0
```
wcole1-godaddy: **Two things here.**

1. **cli v3 exit code vs main's snapshot step.** Verified on this head: `changeset version --snapshot beta` exits 1 with no unreleased changesets (v2.29.5 exits 0). Main's new `pnpm changeset version --snapshot ${{ inputs.tag }}` step will fail on any dispatch when nothing is pending, which is the normal state right after a release. The root `version` script fails the same way locally. This needs to be handled when the conflict with main is resolved, and the `if: github.event_name == 'push'` guard main added to this step must be re-applied or a manual dispatch will run the full publish flow.

2. **GitHub Releases and tag pushes silently re-enabled.** action@v2 with `changeset publish` emits git-tag events via `CHANGESETS_OUTPUT`, so `create-github-releases` / `push-git-tags` (default true) become active again. They have been effectively off since #1205 switched to `pnpm publish -r`. Last GitHub Release is `@godaddy/react@1.0.1` while npm is at 1.0.56; the next Version Packages merge will push one tag and one Release per package. If intended, say so and fix the stale `NPM_TOKEN` note at AGENTS.md:305. If not, set both inputs to `false`.

Minor: the comment says `v2.1.0` but `@v2` resolves to v2.1.2 today, and `github-token: ${{ secrets.GITHUB_TOKEN }}` (line 41) restates the action's default.

packages/eslint-config-godaddy/package.json
```diff
@@ -23,18 +23,21 @@
...
   },
   "dependencies": {
     "@eslint/js": "^10.0.0",
-    "eslint-plugin-jsdoc": "^50.6.9",
+    "eslint-plugin-jsdoc": "^64.3.9",
     "eslint-plugin-json": "^4.0.1",
-    "eslint-plugin-mocha": "^10.5.0",
-    "globals": "^16.0.0"
+    "eslint-plugin-mocha": "^12.0.2",
+    "globals": "^17.12.0"
```
wcole1-godaddy, line 34: **globals 17 drops AudioWorklet globals from `browser`.** v17.0.0's breaking change split `audioWorklet` out of `browser`, so `...globals.browser` no longer declares `AudioWorkletProcessor`, `registerProcessor`, `sampleRate`, `currentFrame`, `currentTime`, `AudioWorkletGlobalScope`, `WorkletGlobalScope`, `Float16Array`. With `js.configs.recommended` (`no-undef: error`) a consumer linting a worklet file goes from passing on 9.0.0 to exit 1.

Add `...globals.audioWorklet` or document the drop in the changeset.

packages/react/package.json
```diff
@@ -123,12 +123,12 @@
...
     "react": "^19",
     "react-dom": "^19",
-    "react-hook-form": "^7.54.2",
+    "react-hook-form": "^7.55.0",
     "tsdown": "^0.15.6",
     "typescript": "~5.7.3",
     "vite": "^6.4.1",
-    "@vitest/coverage-v8": "4.1.2",
-    "vitest": "^4.1.2"
+    "@vitest/coverage-v8": "5.0.0",
```
wcole1-godaddy, line 130: **Exact pin next to a caret.** `@vitest/coverage-v8` is pinned `5.0.0` while `vitest` (line 131) is `^5.0.0`, but coverage-v8@5.0.0 peers on vitest exactly `5.0.0`. Dependabot has already split this pair twice (051ccd5, 8401b9f bumped only vitest), and vitest 5.0.1 is already on the registry. `cicd.yml` only runs `vitest run`, so the mixed-versions warning never shows in CI.

Give both the same specifier, or move them to a pnpm `catalog:` entry. Separately, app-connect defines `test:coverage` but does not depend on the coverage provider, so that script fails under pnpm's isolated node_modules.

packages/react/package.json
```diff
@@ -57,12 +57,12 @@
     "@tanstack/react-query": ">=5",
     "react": "18.x || 19.x",
     "react-dom": "18.x || 19.x",
-    "react-hook-form": ">=7"
+    "react-hook-form": ">=7.55.0"
```
wcole1-godaddy: **Peer wider than the resolver's peer.** `>=7.55.0` advertises react-hook-form 8.x support, but `@hookform/resolvers@5.9.1` (a hard dependency here) peers on `^7.55.0`. Once RHF 8 stable ships (8.0.0-beta.3 today; prereleases do not satisfy the range) a consumer on RHF 8 satisfies this peer and then hits ERESOLVE from inside @godaddy/react's own tree, or runs the resolver helpers against an untested major.

`^7.55.0` is the self-consistent range. The open upper bound is pre-existing; this PR only raised the floor, but it is the natural moment to fix it.

packages/app-connect/package.json
```diff
@@ -61,13 +61,13 @@
 	},
 	"devDependencies": {
 		"@biomejs/biome": "1.9.4",
-		"@changesets/cli": "^2.29.4",
+		"@changesets/cli": "^3.0.2",
```
wcole1-godaddy: **Second `@changesets/cli` and second `release` script.** app-connect keeps its own `@changesets/cli` devDependency (bumped here in lockstep with root) and its own `release: pnpm build && changeset publish` (line 41), a leftover from when it was a standalone repo. `pnpm --filter @godaddy/app-connect release` builds only app-connect and then publishes every unpublished workspace package (verified: `@manypkg/find-root` walks up to the pnpm workspace). With the new root `release` script there are now two entrypoints with different semantics, and AGENTS.md:27 matches neither.

Suggest removing the package-level script and devDependency (`prepublishOnly` already builds) and updating AGENTS.md to the real root script.

wcole1-godaddy (commented):
Review of the consolidated bumps. In-repo checks pass at this head (frozen install, all four eslint-config test suites on ESLint 10.10, app-connect and react on Vitest 5, react `tsc --noEmit`), so the comments below are about published-package consumers, the release path, and semver hygiene.

Two things changed since this was opened:

1. **Conflict with main.** #1464 / #1465 added a `workflow_dispatch` snapshot path to `release.yaml`, and GitHub now reports this PR as CONFLICTING on that file.
2. **@changesets/cli v3 breaks that snapshot path.** Verified on this head: `changeset version` (with or without `--snapshot`) exits 1 when there are no unreleased changesets; v2.29.5 exits 0 on the same tree. Main's `pnpm changeset version --snapshot ${{ inputs.tag }}` step is normally run right after a release, when nothing is pending, so it would fail every time after this merges.

Suggestion: split the changesets/action v2 + cli v3 migration into its own PR rebased on main, where the snapshot path can be adapted deliberately (re-apply `if: github.event_name == 'push'` on the action step, guard or tolerate the exit-1, decide whether snapshot publish stays `pnpm publish -r --tag`). The dependency bumps can then land with the inline fixes below, chiefly dropping the dead Babel parser rather than bumping it, and adding changesets for the peer/engines changes.

Smaller items not anchored to a diff line:
- `.changeset/config.json` `$schema` still points at `@changesets/config@2.0.0`; v3 uses config v4 and removed the `prettier` option.
- Lockfile now has both `globals@17.12.0` and `globals@17.8.0` (the latter via eslint-plugin-mocha). `pnpm dedupe` collapses them.
- `examples/nextjs` still declares `zod ^3.24.1`; `@hookform/resolvers@5` needs ^3.25 and only hoisting keeps it on 3.25.x today.
- `@changesets/cli` 3.0.3 is out with pnpm-10 publish fixes; worth taking when rebasing.

package.json
```diff
@@ -5,7 +5,7 @@
   "description": "The approach to JavaScript at GoDaddy. We think it's pretty decent.",
   "type": "module",
   "engines": {
-    "node": ">=22"
+    "node": "^22.22.2 || ^24.15.0 || >=26"
```
wcole1-godaddy: See the engines comment on `packages/eslint-config-godaddy/package.json:27`. This range excludes all of Node 25.x while the published package's range allows it. None of the deps that forced the bump (jsdoc, mocha, vitest) exclude 25; only @changesets/cli does. If the intent is to mirror jsdoc's floor, copy its string verbatim here too.

package.json
```diff
@@ -41,15 +42,15 @@
     "url": "https://github.com/godaddy/javascript/issues"
   },
   "devDependencies": {
-    "@changesets/cli": "^2.29.5",
+    "@changesets/cli": "^3.0.2",
```
wcole1-godaddy: v3's `changeset version` exits 1 when there are no unreleased changesets (verified on this head; v2 exits 0). That changes the root `version` script from a no-op success to a failure on a clean tree, and breaks main's new `changeset version --snapshot` step. See the comment on `release.yaml:37`. Also: 3.0.3 is out with pnpm-10 publish fixes.

wcole1-godaddy (approved):
Re-reviewed the updated head. Earlier compatibility and snapshot-versioning concerns are addressed; no new actionable findings. Frozen install, full tests/builds, React/localizations typechecks, and app-connect coverage passed. Live npm/OIDC publishing remains unverified.

PR #1467, opened 2026-09-16
remove required GDP appId

pbennett1-godaddy:
## Summary

GoDaddy payment providers and credit card forms were blocked from loading without `godaddyPaymentsConfig.appId`, although the Collect runtime only requires `businessId`.

Remove app ID requirements from SDK loading, provider wrapping, payment method availability, and card, ACH, and wallet initialization. Preserve configured app IDs and session overrides.

## Changeset

Patch changeset for `@godaddy/react`.

## Test Plan

- Added regression coverage for card and ACH mounting with omitted or empty app IDs, session business ID fallback, and missing business IDs.
- Verified wallet initialization without an app ID and preservation of the session app ID override.
- `pnpm --filter @godaddy/react test` — all 596 tests passed.
- `pnpm --filter @godaddy/react typecheck` — passed.
- `pnpm --filter @godaddy/react build` — passed.


packages/react/src/components/checkout/payment/utils/use-load-poynt-collect.ts
```diff
@@ -34,7 +32,6 @@ export function useLoadPoyntCollect() {
     if (
       isPoyntCDNLoaded ||
       isPoyntLoaded ||
-      !applicationId?.trim() ||
       !collectCDN ||
       !businessId
```
wcole1-godaddy, line 41: [P2] Gate Collect loading on a configured GoDaddy payment method

PaymentForm calls this hook unconditionally, and sessions using other processors can still have a businessId. Removing the application ID guard therefore makes Stripe-only and offline-only checkouts download and execute Collect even when godaddyPaymentsConfig is omitted. I reproduced the loader behavior against the base and this commit: neither checkout inserted the script before this change; both insert it afterward. This adds unnecessary network and JavaScript work on the checkout path.

Please gate loading on a configured GoDaddy payment method while keeping appId optional, and add regression coverage showing that non-GoDaddy-only sessions do not insert the script and GoDaddy sessions without an app ID still do.

wcole1-godaddy (changes requested):
Please address the unnecessary Collect SDK load on checkouts that have no GoDaddy payment methods before merging. Keeping application IDs optional makes sense, but the loader should still be gated on whether Collect is needed. The extra download and JavaScript execution add avoidable network and main-thread work to checkout.

packages/react/src/components/checkout/payment/utils/use-load-poynt-collect.ts
pbennett1-godaddy: @wcole1-godaddy - addressed.

wcole1-godaddy (approved):
Reviewed at 162bdd8e30844b684736a81131bca650db4cbd7a. The previously raised Collect SDK-loading concern is addressed: loading is gated on a configured GoDaddy payment method while app IDs remain optional. No actionable findings. Verified all 612 tests, typecheck, and build pass. Live card tokenization and wallet flows were not exercised; payment tests use SDK mocks.

